Issue #11, opened 2021-03-12
K8s object (yaml) for Trap


mayurah:
- UDP Port
- Alignment with poller deployment
- Test in k3s / minikube

Issue #10, opened 2021-03-12
Shell Script to run end to end test


mayurah:
**Based out of Luca's scripts:**
- poller
- mibs
- rabbitmq
- mongodb

**Additions to execute:**
- snmpsim
- snmptrap
- trap server

Issue #9, opened 2021-03-12
Documentation 📝


mayurah:
**Components:**
- Trap
- Poller
- Configuration Scripts
- K8s Environment in Linux

**Sections:**
- [ ]  Prerequisite
      -  Splunk side configrations
             - Splunk Host, HEC Token, SSL Verification, Index Creation
      -  Github token?
- [ ] - Setup K8s Environment (Linux)
       - Install minicube / k3s / git
- [ ] - Configure SC4SNMP
        - Configure Deployments (YAML files)
- [ ] - Deploy SC4SNMP
- [ ] - Test Functionality
- [ ] - Logging (future)
- [ ] - Troubleshooting / FAQ
        -  Login to container
        - Check Health
        - Check Logs
        - Check Network Ports
        - Clean deploy

PR #12, opened 2021-03-12
feat: traps deployment

mayurah:
#9 #10 #11


deploy/sc4snmp/trap/create_deployment.sh
```diff
@@ -0,0 +1,142 @@
...
+k3s kubectl delete service/mib-server-service
+
+health_check
+sleep 5
+
+github_username="${USER}"
+github_email="${github_username}@splunk.com"
+echo "Please type your person access github token:"
+# read -r token
+token=6b4885ba7c27556a7d3d30e06c1ac76843f7cc30
```
lingy1028, line 128: Could you please replace this with a placeholder?

PR #30, opened 2021-04-02
feat: change from Load Balancer to Cluster IP

weliasz:


lstoppa (approved):
Looks fine. These are internal services and should not be available outside the cluster.

PR #75, opened 2021-05-12
Splunk Connect for Kubernetes Configuration Updates

matthewmodestino:
Hey Team! Thanks for all your hard work thus far!

The following PR contains:

- "CLUSTER_NAME" variable typo fix in install_full.bash
- Added SCK global clusterName option rather than set in 3 sub-charts
- SCK scope down logging paths monitored by default to sck and sc4snmp namespaces only. 
- Added reference to sourcetype prefix in case we need it
- SCK disable object and metrics by default
- Reduce default objects polled when enabled and backed off collection to 300s
- Minor config file notes and clean up

Why are these items needed?

Mainly to clean up and slim down the sck components deployed by default to reduce friction to adopt and to ensure we dont overwhelm the user with data not directly related to snmp. We also should be mindful of the worker node VMs with workloads not related to performing snmp tasks. 

As a result of this PR sck deploy will go from 4 containers ( logs, objects, metrics-agg, metrics) down to 1 deployed (logs), and the ingest will be greatly reduced by only looking at the sc4snmp and sck namespaces. 

Users can always explicitly expand the info collected by choice after successfully deploying. I can see metrics moving over to otel, and object remaining a standalone utility Splunk users use.

We ahould also review journald inputs specific to microk8s debug on next iteration.

Tests:

I couldn't get the `install_full.bash` script to point at my local repo updates, so i deployed then deleted the sck helmrelease and re-deployed it manually ensuring the sed commands were enforced.

Deploy resulted in only a single logging pod deployed, and the logs of only the sck and sc4snmp namespace being indexed in Splunk, with the proper cluster name and multiline parsing.


matthewmodestino:
I have read the CLA Document and I hereby sign the CLA

deploy/sc4snmp/sck.yaml
```diff
@@ -47,75 +51,47 @@ spec:
...
-            ]
-        apps:
-          v1:
-            [
-              { "name": "daemon_sets", "interval": "60s" },
-              { "name": "deployments", "interval": "60s" },
-              { "name": "replica_sets", "interval": "60s" },
-              { "name": "stateful_sets", "interval": "60s" },
-            ]
+            [{ "name": "pods", "interval": "300s" },{ "name": "nodes", "interval": "300s" },{ "name": "component_statuses", "interval": "300s" },{ "name": "namespaces", "interval": "300s" },{ "name": "services", "interval": "300s" },{ "name": "events", "mode": "watch" }]
```
lstoppa, line 81: Can you please split this long like like above? This way the configuration would be much more readable.

deploy/sck/sck_147.yaml
```diff
@@ -38,57 +46,50 @@ splunk-kubernetes-logging:
...
   kubernetes:
     insecureSSL: true
-    clusterName: ##CUSTER_NAME##
   objects:
     core:
-      v1: [{"name":"pods", "interval":"60s"},{"name":"nodes", "interval":"60s"},{"name":"component_statuses", "interval":"60s"},{"name":"config_maps", "interval":"60s"},{"name":"namespaces", "interval":"60s"},{"name":"persistent_volumes", "interval":"60s"},{"name":"persistent_volume_claims", "interval":"60s"},{"name":"resource_quotas", "interval":"60s"},{"name":"services", "interval":"60s"},{"name":"service_accounts", "interval":"60s"},{"name":"events", "mode":"watch"},]
-    apps:
       v1:
-        [{"name":"daemon_sets", "interval":"60s"},{"name":"deployments", "interval":"60s"},{"name":"replica_sets", "interval":"60s"},{"name":"stateful_sets", "interval":"60s"},]
+        [{ "name": "pods", "interval": "300s" },{ "name": "nodes", "interval": "300s" },{ "name": "component_statuses", "interval": "300s" },{ "name": "namespaces", "interval": "300s" },{ "name": "services", "interval": "300s" },{ "name": "events", "mode": "watch" }]
```
lstoppa, line 78: Can you please split this long like like my first comment above? This way the configuration would be much more readable.

lstoppa (commented):
The only comment if related to those long lines (see my comments).

matthewmodestino:
Hey @lstoppa I actually liked your formatting, however in my testing I believe it broke my flux deploy when I was working with it. 

While it may be readable, the spacing is tough to work with when crafting custom Configs. If you are not careful how you format the linebreaks, it will break the yaml, like it did in my testing.

How did you create the yaml and ensure no breaks or spaces?

There's also a more readable and k8s yaml safe way to express it and maintain readability...

https://github.com/splunk/splunk-connect-for-kubernetes/blob/eab9ab19617fcd153e47aac417c80727f7edf0e7/helm-chart/splunk-connect-for-kubernetes/values.yaml#L602

lstoppa:
> Hey @lstoppa I actually liked your formatting, however in my testing I believe it broke my flux deploy when I was working with it.
> 
> While it may be readable, the spacing is tough to work with when crafting custom Configs. If you are not careful how you format the linebreaks, it will break the yaml, like it did in my testing.
> 
> How did you create the yaml and ensure no breaks or spaces?
> 
> There's also a more readable and k8s yaml safe way to express it and maintain readability...
> 
> https://github.com/splunk/splunk-connect-for-kubernetes/blob/eab9ab19617fcd153e47aac417c80727f7edf0e7/helm-chart/splunk-connect-for-kubernetes/values.yaml#L602

That was Ryan's magic. if the formatting can easily break the code, then better to leave it the way it is now.

lstoppa (approved):

PR #78, opened 2021-05-25
fix: fixed IPv6 default behavior

lingy1028:


mayurah (approved):
Looks good! Thanks @lingy1028  / @lstoppa for adding a quick fix.

PR #86, opened 2021-08-18
Feature/add performance data

aryznar-splunk:


docs/source/planning.rst
```diff
@@ -35,3 +35,16 @@ Requirements
...
     * Ability to create of obtain real and token
 
+Planning Infrastructure
+===================================================
+
+Single installation of Splunk Connect for SNMP (SC4SNMP) on
+machine with 16 Core/32 threads x64 and 12 GB ram will be able to handle up to 1300 SNMP TRAPs per sec.
+
+Single installation of Splunk Connect for SNMP (SC4SNMP) on
+machine with 16 Core/32 threads x64 and 64 GB ram will be able to handle up to 1300 SNMP GETs per sec.
```
weliasz, line 45: For both 12 GB and 64 GB 1300 SNMP GETs per sec is the limit or is it copy paste bug?

aryznar-splunk: no it is not a but, we tested separately Trap and Poler and both has different memory consumption.

weliasz (approved):

PR #99, opened 2021-09-15
Feat/index change

weliasz:


test/deploy_and_test.sh
```diff
@@ -29,7 +29,7 @@ create_splunk_secret() {
 create_splunk_indexes() {
   splunk_ip=$1
   splunk_password=$2
-  index_names=("em_metrics" "em_meta" "em_logs")
+  index_names=("em_metrics" "em_meta" "em_events")
```
okuzhel-splunk, line 32: With change of the index poller should deal on its own, right? it will see that ocnf is changed and update jobs, correct?

test/otel-collector-config.yaml
```diff
@@ -0,0 +1,21 @@
+receivers:
+  splunk_hec/logs:
```
okuzhel-splunk, line 2: if we changed em_logs to em_events maybe it makes sense to change it here also?

weliasz: This name is internal to OTEL. It can take value of logs, metrics or traces. It has nothing to do with Splunk index names.

test/deploy_and_test.sh
weliasz: During installation of SC4SNMP indexes are not created. This script is solely for purpose of our tests. This was the last part that was skipped during refactoring.

The install.bash script accepts:
    METRICS_INDEX=em_metrics \
    EVENTS_INDEX=em_events \
    META_INDEX=em_meta \
which are later used by our software.

okuzhel-splunk (approved):

okuzhel-splunk (approved):

omrozowicz-splunk (approved):

PR #100, opened 2021-09-20
doc: develop documentation for enricher feature

omrozowicz-splunk:
Documentation for enricher feature


docs/configuration.md
```diff
@@ -303,6 +303,77 @@ profiles:
...
+the **enricher** must be structured as following:
+```yaml
+enricher:
+  oidFamily:
+    IF-MIB:
+      existingVarBinds:
+        - ifIndex: 'interface_index'
+        - ifDescr: 'interface_desc'
+```
+From now on in every query for IF-MIB two new fields "interface_index" and "interface_desc" are visible:
```
okuzhel-splunk, line 324: every query, but what type of query are we talking about? Maybe elaborate a little here

docs/configuration.md
```diff
@@ -303,6 +303,77 @@ profiles:
...
+```yaml
+oid-type1="ObjectIdentity" value1-type="Integer" 1.3.6.1.2.1.2.2.1.7.2="1" value1="1" IF-MIB::ifAdminStatus.2="up" interface_index="2" interface_desc="eth0"
+```
+The value of newly added properties is calculated according to current query index, for IF-MIB::ifAdminStatus.**2** we're interested in IF-MIB::ifIndex.**2** and IF-MIB::ifDescr.**2**.
```
okuzhel-splunk, line 328: I think it would be easier to understadn divided into 2 sentences 

The value of newly added properties is calculated according to current query index. 
For example, for IF-MIB::ifAdminStatus.**2** we're interested in IF-MIB::ifIndex.**2** and IF-MIB::ifDescr.**2**.

okuzhel-splunk (approved):

PR #192, opened 2021-12-01
fix: poller config issue

omrozowicz-splunk:


weliasz (approved):

omrozowicz-splunk:
Tested with Paweł, confirmed that it works

PR #203, opened 2021-12-02
fix: send address of trap origin

weliasz:


splunk_connect_for_snmp/snmp/tasks.py
```diff
@@ -341,8 +341,9 @@ def process_snmp_data(self, varBindTable, metrics, mapping={}):
                     metrics[group_key] = {
                         "metrics": {},
                         "fields": {},
-                        "profiles": set(),
                     }
+                    if len(mapping) > 0:
```
omrozowicz-splunk, line 345: let's use `if mapping:` instead

omrozowicz-splunk (changes requested):

omrozowicz-splunk (approved):

PR #713, opened 2023-03-14
feat: conditional profiles

omrozowicz-splunk:
# Description

This feature enables an option to configure profiles that will be automatically assigned based on a walk data.
Additionally, enriching metrics with indexes was simplified.

Fixes # (issue)

## Type of change

- [x] New feature

## How Has This Been Tested?

Unit tests, manual tests, integration tests

## Checklist


splunk_connect_for_snmp/inventory/tasks.py
```diff
@@ -106,9 +128,18 @@ def generate_poll_task_definition(
     return task_config
 
 
+def add_profile_to_assigned_list(
+    assigned_profiles: dict[int, list[str]], frequency, profile_name
```
mbukowski-splunk, line 132: as frequency seems to be an int, I am wondering whether  it should not be better to use type annotation and frequency: int

splunk_connect_for_snmp/inventory/tasks.py
```diff
@@ -222,3 +255,91 @@ def is_smart_profile_valid(profile_name, profile):
...
+def filter_condition_on_database(mongo_client, address: str, conditions: list):
+    attributes = mongo_client.attributes
+    query = create_query(conditions, address)
+    result = attributes.find(
+        query, {"address": 1, "group_key_hash": 1, "_id": 0, "indexes": 1}
+    )
+    return list(result)
+
+
+def create_profile(profile_name, varBinds, records):
```
mbukowski-splunk, line 269: I am wondering whether:
def create_profile(profile_name, varBinds, records):
    varbind_copy = varBinds.copy()
    indexes_copy = [record["indexes"] for record in records]
    varbind_list = [vb + idx for idx in indexes_copy for vb in varbind_copy]
    profile = {profile_name: {"frequency": 10, "varBinds": varbind_list}}
    return profile
is a bit more optimal:
1. list comprehension instead of for loop. it is faster
2. dictionary compehension instead of empty dict and then adding values. it should be a bit faster as well
3. In the current inner loop the varbind var is copy in each iteration of this loop. I think it will be faster if we copy this outisde of the loop.

What do you think?

splunk_connect_for_snmp/inventory/tasks.py
```diff
@@ -222,3 +255,91 @@ def is_smart_profile_valid(profile_name, profile):
...
+        "in": "$in",
+    }
+
+    def _parse_mib_component(field: str) -> str:
+        mib_component = field.split("|")
+        if len(mib_component) < 2:
+            raise BadlyFormattedFieldError(f"Field {field} is badly formatted")
+        return mib_component[0]
+
+    def _convert_to_float(value: str, ignore_error=False) -> typing.Any:
```
mbukowski-splunk, line 293: This is actually nice implementation.I would only check before the conversion to float(value) whether it is numeric:
if(value.isnumeric())
   return float(value)
you can also use try-except-else block if you want to avoid catching the ValueError exception.

mbukowski-splunk (changes requested):

splunk_connect_for_snmp/inventory/tasks.py
omrozowicz-splunk: I use copy of varBinds element, not the varBinds itself, so can't copy it before the loop. But, you're right I can use list comprehension because I don't even remember why the copy operator is here - probably I used it before in some other way and forgot to delete it later on. Also, you found out how I terribly hardcoded frequency 🙈 Thanks!

splunk_connect_for_snmp/inventory/tasks.py
omrozowicz-splunk: Yes, but here I have two cases - in one I should ignore that the value can't be float (when there is an in operator that can have any value) and the one where I should inform user that he configured profile wrongly - when he used lt or gt that can be only numeric. Not sure how try-except-else block would help me in that. I wouldn't go for the `isnumeric()` option as it has many limitations - for example it won't work with decimal points. For ex:
```
>>> "1.2".isnumeric()
False
```

PR #714, opened 2023-03-20
fix: do not poll IF-MIB by default

omrozowicz-splunk:
# Description

This PR limits the default small walk to not include `IF-MIB` family as this is not what we always want to poll from the devices.
Additionally it adds a way to stop polling base profiles as not everyone need them.

Fixes # (issue)

## Type of change

- [x] Refactor/improvement

## How Has This Been Tested?

## Checklist

### DIFF
--- a/docs/configuration/configuring-profiles.md
+++ b/docs/configuration/configuring-profiles.md
@@ -160,8 +160,8 @@ poller:
 ```
 
 NOTE: When small walk is configured, you can set up polling only of OIDs belonging to the walk profile varBinds. 
-Additionally, there are two MIB families that are enabled by default (we need them to create the state of the device in the database and poll base profiles): `IF-MIB` and `SNMPv2-MIB`.
-For example, if you've decided to use `small_walk` from the example above, you'll be able to poll only `UDP-MIB`, `IF-MIB`, and `SNMPv2-MIB` OIDs.
+Additionally, `SNMPv2-MIB` that is enabled by default (we need it to create the state of the device in the database).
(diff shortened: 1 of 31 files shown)


docs/configuration/configuring-profiles.md
```diff
@@ -160,7 +160,7 @@ poller:
 ```
 
 NOTE: When small walk is configured, you can set up polling only of OIDs belonging to the walk profile varBinds. 
-Additionally, there are two MIB families that are enabled by default (we need them to create the state of the device in the database and poll base profiles): `IF-MIB` and `SNMPv2-MIB`.
+Additionally, `SNMPv2-MIB` that is enabled by default (we need them it create the state of the device in the database).
```
wojtekzyla, line 163: "we need it to create"

docs/configuration/configuring-profiles.md
```diff
@@ -160,7 +160,7 @@ poller:
...
 For example, if you've decided to use `small_walk` from the example above, you'll be able to poll only `UDP-MIB`, `IF-MIB`, and `SNMPv2-MIB` OIDs.
```
wojtekzyla, line 164: shouldn't it be "poll only `UDP-MIB` and `SNMPv2-MIB` OIDs."

docs/configuration/configuring-profiles.md
omrozowicz-splunk: fixed

docs/configuration/configuring-profiles.md
omrozowicz-splunk: right, changed

PR #806, opened 2023-08-03
chore: traps aggregation

ajasnosz:
# Description

Adds traps aggregation into one event if flag aggregateTrapsEvents is set to true.

Fixes # (issue)

## Type of change

- [x] New feature

## How Has This Been Tested?

unit test, manual test

## Checklist


splunk_connect_for_snmp/splunk/tasks.py
```diff
@@ -238,8 +239,19 @@ def prepare_trap_data(work):
             "host": work["address"],
             "index": SPLUNK_HEC_INDEX_EVENTS,
         }
-        events.append(json.dumps(event, indent=None))
-
+        if SPLUNK_CONGREGATE_TRAPS_EVENTS:
```
omrozowicz-splunk, line 242: I think I'd create a new function here for this if statement. Or I'm thinking - if we just do `events.append` for both, and run `json.dumps` at the end - won't it return the same result?

omrozowicz-splunk (changes requested):

splunk_connect_for_snmp/splunk/tasks.py
ajasnosz: I'll add new function. For events I think dumps can be done at the end if I add the loop.

docs/configuration/trap-configuration.md
```diff
@@ -136,3 +136,11 @@ traps:
     annotations:
       annotation_key: annotation_value
 ```
+
+### Aggregate traps
+In case you want to see traps events collected as one event inside splunk you can enable it by setting `splunk.aggregateTrapsEvents`.
+Example:
+```yaml
+splunk:
```
omrozowicz-splunk, line 144: I think it is better to keep this under `traps` section instead of `splunk`, and even here in docs is under traps' configuration

test/splunk/test_prepare.py
```diff
@@ -73,6 +73,62 @@ def test_prepare_trap(self, m_custom):
...
+        m_custom.return_value = task_input
+        result = prepare(task_input)
+
+        self.assertEqual("list", type(result["events"]).__name__)
+        self.assertEqual(2, len(result))
+        self.assertEqual(1, len(result["events"]))
+        self.assertEqual(0, len(result["metrics"]))
+
+        item1 = json.loads(result["events"][0])
+        event1 = json.loads(item1["event"])
```
omrozowicz-splunk, line 116: Can we change it into a real json object so it's easy to see how the final structure looks like?

omrozowicz-splunk (changes requested):

docs/configuration/trap-configuration.md
ajasnosz: done

test/splunk/test_prepare.py
ajasnosz: done

docs/configuration/trap-configuration.md
omrozowicz-splunk: don't you need a one newline before yaml section? does it render properly ?

ajasnosz: it renders normally
image

PR #807, opened 2023-08-03
fix: while configuring host inside the group without explicitly specifying its port, and later adding explicite port, the previous configuration with the default port from the inventory will be deleted

wojtekzyla:
# Description

While configuring host inside the group without explicitly specifying its port, and later adding explicite port, the previous configuration with the default port from the inventory will be deleted

## Type of change

- [x] Bug fix

## How Has This Been Tested?

Added and updated existing unit tests

## Checklist

### DIFF
--- a/splunk_connect_for_snmp/common/inventory_processor.py
+++ b/splunk_connect_for_snmp/common/inventory_processor.py
@@ -57,22 +57,29 @@ def gen_walk_task(ir: InventoryRecord, profile=None, group=None):
     return task_config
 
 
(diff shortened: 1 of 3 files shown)


omrozowicz-splunk (commented):
I'm thinking if instead of this we can use the same notation that we use for hosts, so port is a part of the key.
Like in case of 161, the group key would be `group`, but in case of 162 it will be `group:162`.
Then we won't need to add whole this logic and only manipulate the key instead. What do you think?

omrozowicz-splunk (approved):

PR #838, opened 2023-08-11
Fix/expires timeout configuration

wojtekzyla:
# Description

Add configuration of tasks expiration time

- [x] Refactor/improvement

## How Has This Been Tested?

Manual and usnit testing

## Checklist

### DIFF
--- a/charts/splunk-connect-for-snmp/values.yaml
+++ b/charts/splunk-connect-for-snmp/values.yaml
@@ -175,6 +175,9 @@ scheduler:
   podAntiAffinity: soft
   nodeSelector: {}
 
+  # tasks expirations time in seconds
+  tasksExpiryTime: 60
+
 poller:
   # Appending OID indexes to metrics.
   # https://splunk.github.io/splunk-connect-for-snmp/main/configuration/poller-configuration/#append-oid-index-part-to-the-metrics
--- a/docs/configuration/scheduler-configuration.md
+++ b/docs/configuration/scheduler-configuration.md
@@ -69,3 +69,11 @@ scheduler:
 The one obligatory field for the host configuration is `address`. If `port` isn't configured its default value is `161`. 
 Other fields that can be modified here are: `community`, `secret`, `version`, `security_engine`.
 However, if they remain unspecified in the host configuration, they will be derived from the inventory record regarding this specific group.
+
+### Define tasks expiry time
+
+Define time in second after which polling or walk tasks, that haven't been picked up by the worker, will be revoked. Check the [celery documentation](https://docs.celeryq.dev/en/stable/userguide/calling.html#expiration) for more details.
+```yaml
+scheduler:
+  tasksExpiryTime: 300
+```
--- a/splunk_connect_for_snmp/common/task_generator.py
+++ b/splunk_connect_for_snmp/common/task_generator.py
@@ -1,9 +1,19 @@
+import os
+
 from celery import Celery, chain, group, signature
 from celery.schedules import schedule
 
+CHAIN_OF_TASKS_EXPIRY_TIME = int(os.getenv("CHAIN_OF_TASKS_EXPIRY_TIME", "60"))
+
 
 class TaskGenerator:
-    def __init__(self, target: str, schedule_period: int, app: Celery, host_group=None):
+    def __init__(
+        self,
+        target: str,
+        schedule_period: int,
+        app: Celery,
+        host_group=None,
+    ):
         self.target = target
         self.schedule_period = schedule_period
         self.app = app
@@ -24,8 +34,8 @@ class TaskGenerator:
 
 
 class WalkTaskGenerator(TaskGenerator):
-
     WALK_CHAIN_OF_TASKS = {
+        "expires": CHAIN_OF_TASKS_EXPIRY_TIME,
         "link": chain(
             signature("splunk_connect_for_snmp.enrich.tasks.enrich")
             .set(queue="poll")
@@ -48,7 +58,14 @@ class WalkTaskGenerator(TaskGenerator):
         ),
     }
 
-    def __init__(self, target, schedule_period, app, host_group, profile):
+    def __init__(
+        self,
+        target,
+        schedule_period,
+        app,
+        host_group,
+        profile,
+    ):
         super().__init__(target, schedule_period, app, host_group)
         self.profile = profile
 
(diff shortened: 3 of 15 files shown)


charts/splunk-connect-for-snmp/values.yaml
```diff
@@ -175,6 +175,9 @@ scheduler:
   podAntiAffinity: soft
   nodeSelector: {}
 
+  # tasks expirations time in seconds
+  tasksExpiryTime: 180
```
omrozowicz-splunk, line 179: maybe let's not change the default one, I think the default one was 60 seconds

docs/configuration/scheduler-configuration.md
```diff
@@ -69,3 +69,11 @@ scheduler:
 The one obligatory field for the host configuration is `address`. If `port` isn't configured its default value is `161`. 
 Other fields that can be modified here are: `community`, `secret`, `version`, `security_engine`.
 However, if they remain unspecified in the host configuration, they will be derived from the inventory record regarding this specific group.
+
+### Define tasks expiry time
+
+Define time in second after which polling or walk tasks, that haven't been picked up by the worker, will be revoked.
```
omrozowicz-splunk, line 75: maybe let's add here the page from celery documentation?

splunk_connect_for_snmp/common/task_generator.py
```diff
@@ -24,32 +32,44 @@ def generate_task_definition(self):
...
-                    .set(priority=1),
-                    signature("splunk_connect_for_snmp.splunk.tasks.send")
-                    .set(queue="send")
-                    .set(priority=0),
+                .set(priority=4),
+                group(
+                    signature(
+                        "splunk_connect_for_snmp.inventory.tasks.inventory_setup_poller",
+                        kwargs={
+                            "chain_of_tasks_expiry_time": self.chain_of_tasks_expiry_time
```
omrozowicz-splunk, line 57: why this parameter is also in kwargs?

omrozowicz-splunk (changes requested):
Tbh, I think it might have worked better if we just sense if value of expires is different than in the previous run in migration function, and if so - we delete all the tasks and set up new ones

wojtekzyla:
> > Tbh, I think it might have worked better if we just sense if value of expires is different than in the previous run in migration function, and if so - we delete all the tasks and set up new ones
> 
> like here:
> 
> https://github.com/splunk/splunk-connect-for-snmp/blob/1a013ea0b8d9ae4483bcdb22e24442b09cc127a1/splunk_connect_for_snmp/common/schema_migration.py#L79

Sure, however we also need to delete all existing walk tasks, as we are using `expires` variable  also in those tasks.

omrozowicz-splunk:
> 



> > > Tbh, I think it might have worked better if we just sense if value of expires is different than in the previous run in migration function, and if so - we delete all the tasks and set up new ones
> > 
> > 
> > like here:
> > https://github.com/splunk/splunk-connect-for-snmp/blob/1a013ea0b8d9ae4483bcdb22e24442b09cc127a1/splunk_connect_for_snmp/common/schema_migration.py#L79
> 
> Sure, however we also need to delete all existing walk tasks, as we are using `expires` variable also in those tasks.

Yes, I think it's better to recreate the walks instead of providing everywhere that complicated logic

splunk_connect_for_snmp/common/task_generator.py
wojtekzyla: This is because polling tasks are set up inside walk tasks. We need to somehow pass this parameter to the `inventory_setup_poller ` function, so that worker running this walk task knows what should be the `expiry` value for polling tasks.

I also tested the solution when we don't pass this parameter this way, instead we take it from the environmental variable inside a worker which sets up new polling tasks. The idea was that after this env variable is changed, the worker would restart with the new env variable value and new polling tasks would be created with updated `expiry` time. The problem was, that after running `helm upgrade` with env variable changed, that old worker still worked for a short period of time, and some of the records from the inventory didn't have `expiry` value changed while others had. 
So I decided that we will be checking `CHAIN_OF_TASKS_EXPIRY_TIME` env variable only in `loader` where it is always going to be up to date and we will pass this value to walk tasks. New polling tasks are later created inside walk task by the workers.

splunk_connect_for_snmp/common/schema_migration.py
```diff
@@ -67,10 +67,29 @@ def migrate_database(mongo_client, periodic_obj):
...
 
         save_schema_version(mongo_client, CURRENT_SCHEMA_VERSION)
 
+    previous_expiry_time = periodic_obj.get_chain_of_task_expiry()
+    logger.info(
+        f"Chain of tasks expiry time. Previous: {previous_expiry_time}. Current: {chain_of_tasks_expiry_time}"
+    )
+    expiry_time_changed = False
+    if (
+        previous_expiry_time is None
```
omrozowicz-splunk, line 85: why it is done as when previous expiry time is None? that indicates there's no tasks yet

splunk_connect_for_snmp/common/task_generator.py
omrozowicz-splunk: after the newest changes - assuming we're always deleting all the polls and walks, can we just use env variable to set it up?

omrozowicz-splunk (changes requested):
Ok, I was thinking about it one more time and I'd even simplify it even more. We can either use that get expire from current tasks from redis in every load, and if it's different -> delete walks and polls and add new ones and retrigger, or (I think better idea) trigger another check similar to migrate database, keep one record in the db and check if it has changed - if yes, delete walks, polls and recreate and retrigger walks. What do you think?

splunk_connect_for_snmp/common/task_generator.py
```diff
@@ -24,15 +34,15 @@ def generate_task_definition(self):
...
     WALK_CHAIN_OF_TASKS = {
+        "expires": CHAIN_OF_TASKS_EXPIRY_TIME,
         "link": chain(
             signature("splunk_connect_for_snmp.enrich.tasks.enrich")
             .set(queue="poll")
             .set(priority=4),
             group(
                 signature(
-                    "splunk_connect_for_snmp.inventory.tasks.inventory_setup_poller"
+                    "splunk_connect_for_snmp.inventory.tasks.inventory_setup_poller",
```
omrozowicz-splunk, line 45: was this comma provided by pre-commit?

test/inventory/test_loader.py
```diff
@@ -98,14 +141,26 @@ def test_walk_task_for_port_161(self):
...
         self.assertEqual("sc4snmp;192.68.0.1;walk", result["name"])
         self.assertEqual("splunk_connect_for_snmp.snmp.tasks.walk", result["task"])
         self.assertEqual("192.68.0.1", result["target"])
         self.assertEqual([], result["args"])
-        self.assertEqual({"address": "192.68.0.1", "profile": None}, result["kwargs"])
+        self.assertEqual(
+            {
+                "address": "192.68.0.1",
+                "profile": None,
+                "chain_of_tasks_expiry_time": chain_of_tasks_expiry_time,
```
omrozowicz-splunk, line 155: Can you add a test when `chain_of_tasks_expiry_time` is `None`?

wojtekzyla: Here `chain_of_tasks_expiry_time ` will never be `None` because it is created [here](https://github.com/splunk/splunk-connect-for-snmp/blob/fix/expires-timeout-configuration/splunk_connect_for_snmp/common/task_generator.py#L81), where this variable is set from env variable which has default value of 60 ([here](https://github.com/splunk/splunk-connect-for-snmp/blob/fix/expires-timeout-configuration/splunk_connect_for_snmp/common/task_generator.py#L6C14-L6C14).

However, there is one function which returns `None` if there are no tasks in redbeat or if existing tasks don't have an `expires` variable and this is [get_chain_of_task_expiry](https://github.com/splunk/splunk-connect-for-snmp/blob/fix/expires-timeout-configuration/splunk_connect_for_snmp/customtaskmanager.py#L118-L123). This function is used [here](https://github.com/splunk/splunk-connect-for-snmp/blob/fix/expires-timeout-configuration/splunk_connect_for_snmp/customtaskmanager.py#L55) inside `did_expiry_time_change`. Unit tests of the `get_chain_of_task_expiry` can be found [here](https://github.com/splunk/splunk-connect-for-snmp/blob/fix/expires-timeout-configuration/test/other/test_custom_task_manager.py#L234-L256) and tests if the `did_expiry_time_change` where `did_expiry_time_change` will return `None` is [here](https://github.com/splunk/splunk-connect-for-snmp/blob/fix/expires-timeout-configuration/test/other/test_custom_task_manager.py#L292-L298)

omrozowicz-splunk (approved):

PR #952, opened 2024-02-01
feat: build sc4snmp with gopoller and add logger configuration in go code

wojtekzyla:
# Description

This PR updates the Dockerfile in order to build sc4snmp with a new poller written in GO. This PR also adds configuration of logger in the new GO poller.

## Type of change

- [ ] New feature

## How Has This Been Tested?

Manual testing

## Checklist

### DIFF
--- a/charts/splunk-connect-for-snmp/values.yaml
+++ b/charts/splunk-connect-for-snmp/values.yaml
@@ -239,6 +239,9 @@ worker:
 
   # The poller worker consumes all the tasks related to polling
   poller:
+    go:
+      # Enable logs from gosnmp library
+      gosnmpLogger: false
     # number of the poller replicas when autoscaling is set to false
     replicaCount: 2
     # minimum number of threads in a pod
--- a/splunk_connect_for_snmp/gopoller/auth.go
+++ b/splunk_connect_for_snmp/gopoller/auth.go
@@ -3,6 +3,8 @@ package gopoller
 import (
 	"fmt"
 	"github.com/gosnmp/gosnmp"
+	"log"
+	"os"
 	"time"
 )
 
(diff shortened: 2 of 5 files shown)


splunk_connect_for_snmp/gopoller/auth.go
```diff
@@ -75,6 +77,10 @@ func getSecurityParameters(authData SnmpV3StringAuthData) (*gosnmp.UsmSecurityPa
 		PrivacyPassphrase:        authData.PrivacyPassphrase,
 		AuthoritativeEngineID:    authData.AuthoritativeEngineID,
 	}
+	useLogger := os.Getenv("USE_GO_LOGGER")
```
ajasnosz, line 80: maybe we could name it `ENABLE_GO_LOGGER` or something similar? 
Is it possible to set log level for this logger?
Could we add setting this flag to yaml? I think we could do section for go under
`poller.go` or `worker.poller.go`

ajasnosz (changes requested):

splunk_connect_for_snmp/gopoller/auth.go
wojtekzyla: I can change the name to `ENABLE_GO_LOGGER` and add the configuration to values.yaml. When it comes to the log level, I found the following info in the gosnmp documentation: 
Screenshot 2024-02-01 at 13 51 26

I also checked how this logger is used in code of gosnmp, and they don't distinguish any logging levels. They just simply log if the logger was configured

charts/splunk-connect-for-snmp/values.yaml
```diff
@@ -226,6 +226,10 @@ poller:
   # https://splunk.github.io/splunk-connect-for-snmp/main/configuration/poller-configuration/#define-usernamesecrets
   usernameSecrets: []
 
+  go:
+    # Enable logs from gosnmp library
+    gosnmpLogger: false
```
ajasnosz, line 231: probably we should add it also here https://github.com/splunk/splunk-connect-for-snmp/blob/main/charts/splunk-connect-for-snmp/templates/worker/_helpers.tpl#L81 as env variable in workers, and map this param to ENABLE_GO_LOGGER and I think we can set the default as false.

ajasnosz (changes requested):

splunk_connect_for_snmp/gopoller/auth.go
ajasnosz: Ok, so if it logs everything, we'll follow with that.

charts/splunk-connect-for-snmp/values.yaml
wojtekzyla: Done

ajasnosz (approved):

PR #1072, opened 2024-09-06
chore: update pre-commit config

ajasnosz:
# Description

Fix of failing pre-commit workflow

Fixes # (issue)

## Type of change

- [x] Dependency update
- [x] Bug fix

## How Has This Been Tested?

github actions

### DIFF
--- a/splunk_connect_for_snmp/snmp/manager.py
+++ b/splunk_connect_for_snmp/snmp/manager.py
@@ -294,7 +294,7 @@ class Poller(Task):
         self,
         ir: InventoryRecord,
         walk: bool = False,
-        profiles: List[str] = None,
+        profiles: Union[List[str], None] = None,
     ):
         retry = False
         address = transform_address_to_key(ir.address, ir.port)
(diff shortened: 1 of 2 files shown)


splunk_connect_for_snmp/snmp/manager.py
```diff
@@ -294,7 +294,7 @@ def do_work(
         self,
         ir: InventoryRecord,
         walk: bool = False,
-        profiles: List[str] = None,
+        profiles: List[str] = [],
```
ikheifets-splunk, line 297: I not sure that empty list as default value good idea:
* https://florimond.dev/en/posts/2018/08/python-mutable-defaults-are-the-source-of-all-evil
* https://stackoverflow.com/questions/26320899/why-is-the-empty-dictionary-a-dangerous-default-value-in-python


Can we use `profiles: List[str] = None` and add to body function:


```
if profiles is None:
   ...
```

ikheifets-splunk (approved):

splunk_connect_for_snmp/snmp/manager.py
ajasnosz: changed

PR #1124, opened 2024-11-06
Fix: refactor gather elements

ajasnosz:
# Description

Refactor of gather_elements() method.

Fixes # (issue)

## Type of change

- [x] Refactor/improvement

## How Has This Been Tested?

manual, unit, integration

## Checklist

### DIFF
--- a/splunk_connect_for_snmp/common/collection_manager.py
+++ b/splunk_connect_for_snmp/common/collection_manager.py
@@ -181,36 +181,39 @@ class ProfilesManager(CollectionManager):
                     for key, profile in profiles.items():
                         active_profiles[key] = profile
         if CONFIG_FROM_MONGO:
-            profiles_list = list(self.mongo.sc4snmp.profiles_ui.find({}, {"_id": 0}))
-            for pr in profiles_list:
-                key = list(pr.keys())[0]
-                profile = pr[key]
-                if key in active_profiles:
-                    if not profile.get("enabled", True):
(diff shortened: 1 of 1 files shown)


splunk_connect_for_snmp/common/collection_manager.py
```diff
@@ -181,36 +181,43 @@ def gather_elements(self):
...
-                                if not profile.get("enabled", True):
-                                    celery_logger.info(f"disabling profile {key}")
-                                    del active_profiles[key]
-                                else:
-                                    active_profiles[key] = profile
+            self.gather_profiles_from_config_file(active_profiles)
+        return active_profiles
+
+    @staticmethod
+    def gather_profiles_from_config_file(active_profiles):
```
ikheifets-splunk, line 190: probably better would be name it `merge_profiles_*`, because you we have `active_profiles` that mutable variable and it's super unclear that we modifying it inside function and after that returning it on top level.

It seems this peace of code we using on both functions (super similar) can we reuse it? And I think `merge_profiles` would be good name for it.

```
  for key, profile in profiles.items():
         if key in active_profiles:
             if not profile.get("enabled", True):
                 celery_logger.info(f"disabling profile {key}")
                 del active_profiles[key]
             else:
                 active_profiles[key] = profile
         else:
             active_profiles[key] = profile
```

ajasnosz: I've extracted the common part to new method, but I wouldn't try to merge them into one

PR #1143, opened 2024-12-18
fix: same trap ports

ajasnosz:
# Description

Redesigns the configuration for traps server to receive both IPv4 and Ipv6 traps on the same port. 

Fixes # (issue)

## Type of change

- [x] Refactor/improvement
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## How Has This Been Tested?

manual, integration, unit

## Checklist


splunk_connect_for_snmp/traps.py
```diff
@@ -174,18 +174,19 @@ def main():
...
-        udp.domainName,
-        udp.UdpTransport().openServerMode(("0.0.0.0", 2162)),
-    )
-
+    # UDP socket over IPv6 listens also for IPv4
     if IPv6_ENABLED:
         config.addTransport(
             snmp_engine,
             udp6.domainName,
-            udp6.Udp6Transport().openServerMode(("::", 2163)),
```
ikheifets-splunk, line 188: Probably we can listen both ports 2162 and 2163, **just for backward compatibility**. It would be undocumented.  Because  when users will update SC4SNMP to latest version they will have a problem.

PR #1180, opened 2025-05-19
fix: update mongodb for microk8s

omrozowicz-splunk:
# Description

This PR introduces upgrade of MongoDB from chart version 13.x to 15.x, which is connected with increasing feature compatibility version from 5.0 to 6.0.

The change required adding pre-upgrade job that runs setting compatibility version to `6.0` before upgrading to newer mongo.

+ minor update of mibserver

Fixes # (issue)

## Type of change

Please delete options that are not relevant.

- [x] Dependency update
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] This change requires a documentation update

## How Has This Been Tested?

I dumped the data from `attributes`, `targets` and `inventory` collections before upgrade and compared with the data after the upgrade - it is the same (excluding timestamps from SNMP data), also I manually run the upgrade for a few times. 

Also I'll run integration tests.

## Checklist

- [x] My commit message is [conventional](https://www.conventionalcommits.org/en/v1.0.0/#summary)
- [x] I have run pre-commit on all files before creating the PR
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] I have checked my code and corrected any misspellings

PR #1182, opened 2025-05-20
fix: update mongo docker compose

omrozowicz-splunk:
# Description

This PR introduces an update of mongodb to match updated microk8s mongodb helm version from https://github.com/splunk/splunk-connect-for-snmp/pull/1180

Unlike in microk8s there's no mechanism of upgrade, as we don't specify any in the documentation. 
There's no preffered way of upgrading packages in docker-compose

Fixes # (issue)

## Type of change

- [ ] Dependency update
- [ ] Bug fix
- [ ] New feature
- [ ] Refactor/improvement

## How Has This Been Tested?

## Checklist


ajasnosz (approved):
Update Changelog

omrozowicz-splunk:
> Update Changelog

@ajasnosz done

docker_compose/.env
```diff
@@ -24,7 +24,7 @@ MIBSERVER_TAG=latest
 REDIS_IMAGE=docker.io/bitnami/redis
 REDIS_TAG=7.2.1-debian-11-r0
 MONGO_IMAGE=docker.io/bitnami/mongodb
-MONGO_TAG=6.0.9-debian-11-r5
```
ikheifets-splunk, line 27: probably we can use mongo without Debian? because we migrated sc4snmp base image from debian to alpine only because it had lots of CVE :)

omrozowicz-splunk: we use exactly this image in microk8s subchart, this is the line where it is referenced: https://github.com/bitnami/charts/blob/ded7e1f62ddcaba8fdeaf29a485aaf199295ccfc/bitnami/mongodb/Chart.yaml#L11

It probably should be possible to change the image to a different one in chart, but that might come with additional problems and we'd need to perform a lot of tests to make sure it works fine 😅 I put this tag as I wanted to be consistent with what we have in microk8s. Anyway soon we will migrate to the newest mongo image (but probably still a debian :<)

ikheifets-splunk: ok, thanks

PR #1184, opened 2025-05-22
fix: Fixed the sim pod creation issue when sim.secret.name was set

cwadhwani-splunk:
# Description

Issue:
This issue occurs when both .Values.secret.create and .[Values.secret.name](http://values.secret.name/) are set in the Helm values.
The problem arises because during the secret creation process, the value of .[Values.secret.name](http://values.secret.name/) is not being considered — the secret is created with a default name. However, when the deployment tries to reference the secret, it uses the name from .[Values.secret.name](http://values.secret.name/) (if it's set).
As a result, the deployment references a secret that doesn't exist, which leads to the pod failing to start with a CreateContainerConfigError.
To avoid this issue, we need to ensure that if .[Values.secret.name](http://values.secret.name/) is set, it's also used during secret creation.

Fixes # (issue)

## Type of change

- [x] Bug fix

## How Has This Been Tested?

Manual testing

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - migrate MongoDB subchart to 15.6.26
 
 ### Fixed
+- Ensured `.Vaues.secret.name` is used during secret creation to prevent mismatched secret references in deployments
 
 
 ## [1.12.2]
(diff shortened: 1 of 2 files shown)


omrozowicz-splunk (changes requested):
Hey, lgtm, can you update the PR template with description of why this fix is needed? thanks!

ajasnosz (commented):
Also good with me, please update Changelog.md before merging

cwadhwani-splunk:
@omrozowicz-splunk @ajasnosz 
Added the description in the comment and updated the CHANGELOG.md file.

Issue #1210, opened 2025-07-01
SNMP is reporting trap is report a value as infinity which is not valid json


thel1988:
We have been testing sc4snmp for a while now, we have seen if a device is report values as infinty eg:
```memoryDeviceSerialNumberName": Infinity``` or  ```"fruSerialNumberName": Infinity``` it will cause both the Otel collector and splunk enterprise to not parse it as valid json.

from the JSON spec: 
Infinity is not a valid value in the official JSON specification (RFC 8259). While it's a recognized special numeric value in languages like Python and JavaScript, a compliant JSON parser must reject it. The value should be represented as null (if it's unknown or not applicable) or as a quoted string "Infinity"

Is it possible to overwrite the infinity values with quotes when returned in both snmp traps and snmp walks?

Both a Splunk HEC and a Splunk HEC endpoint hosted on a otel collector is rejecting the events as they should?

**Comment by omrozowicz-splunk** [comment]
Hey @thel1988, Can you show us the snippet of how this `fruSerialNumberName` and/or `memoryDeviceSerialNumberName` value look when it is retrieved by the bare `snmpget`?

We’ll take this into account while planning the work for the upcoming sprints.

**Comment by thel1988** [comment]
Yes ofcause I can

But they return about the same value:
```
IDRAC-MIB-SMIv2::fruSerialNumberName.1.8 = STRING: "2606123B"
IDRAC-MIB-SMIv2::fruSerialNumberName.1.9 = STRING: "2605C123"
```
```
IDRAC-MIB-SMIv2::memoryDeviceSerialNumberName.1.8 = STRING: "2605F3D0"
IDRAC-MIB-SMIv2::memoryDeviceSerialNumberName.1.9 = STRING: "2605F123"
```
No value is out of the ordinary. The weird thing is that, it seems to be only one specific host. Cant really see why it would get the value infinity instead of the serial

**Comment by omrozowicz-splunk** [comment]
Okay, I think know what can be the issue here. Are the hosts returning that Infinity value by any chance returning the strings with only numbers and "E" letters? because we check the type with this function:

```
def value_as_best(value) -> Union[str, float]:
    try:
        return float(value)
    except ValueError:
        return value
    except TypeError:
        return value
```

and when the string looks like this `2605F3D0` then it works as expected, but when it includes only `E` letters it treats it like a mathematical E notation. An example:

```
>>> float("2605F3D0")
Traceback (most recent call last):
[...]

PR #1219, opened 2025-07-23
fix: problem with casting strings of scientific notation

omrozowicz-splunk:
# Description

In SC4SNMP we check the type of SNMP datapoint with this function:

```
def value_as_best(value) -> Union[str, float]:
    try:
        return float(value)
    except ValueError:
        return value
    except TypeError:
        return value
```

and when the string looks like this `2605F3D0` then it works as expected, but when it includes only `E` letters it treats it like a mathematical E notation. An example:

```
>>> float("2605F3D0")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: could not convert string to float: '2605F3D0'
>>> float("260E3100")
inf
```

Also even when the value is not as big to be an infinity, like for ex. `849867E3` it will cast it to be a metric (while it should be a string serial number). 

As far as I know SNMP OID values are typically represented as integers, strings, IP addresses, or octet strings, and the standard SNMP data types such as counter32, gauge, integer, unsigned32, and string do not natively use E notation for their values. For example, numeric values are usually represented as 32-bit integers or unsigned integers without scientific notation formatting.

Fixes # [(issue)](https://github.com/splunk/splunk-connect-for-snmp/issues/1210)

## Type of change

- [x] Bug fix
- [x] Refactor/improvement

## How Has This Been Tested?

Unit tests

## Checklist


test/common/test_humanbool.py
```diff
@@ -52,3 +52,8 @@ def test_convert_to_float_error(self):
         with self.assertRaises(BadlyFormattedFieldError) as context:
             convert_to_float(value)
         self.assertEqual("Value 'up' should be numeric", context.exception.args[0])
+
+    def test_convert_to_float_scientific_notation(self):
+        self.assertEqual(convert_to_float("260E3100"), "260E3100")
+        self.assertEqual(convert_to_float("849867E3"), "849867E3")
+        self.assertEqual(convert_to_float("849867E2"), "849867E2")
```
ikheifets-splunk, line 59: probably we need to check in tests `hex` numbers

ikheifets-splunk (approved):

PR #1224, opened 2025-08-06
feat: added configurable parameter for SNMP request's retry and timeo…

cwadhwani-splunk:
# Description
Added configurable parameter for SNMP request's retry and timeout. Handled tasktimeout configurable parameter for discovery

- Used the existing UDP_CONNECTION_TIMEOUT for snmp requests made by discovery worker.
- Introduced a new env UDP_CONNECTION_RETRIES that will used for all (including poller) the snmp requests.
- Handled tasktimeout for discovery task. Note: This is currently not introduced for other tasks.
- Handled race condition (using a lock mechanism) when discovery tasks tries to update the output csv file concurrently

Fixes # (issue)

## Type of change

- [ ] Dependency update
- [x] Bug fix
- [x] New feature
- [ ] Refactor/improvement

## How Has This Been Tested?

## Checklist


charts/splunk-connect-for-snmp/values.yaml
```diff
@@ -410,6 +410,8 @@ worker:
   podAntiAffinity: soft
   # udpConnectionTimeout timeout in seconds for SNMP operations
   udpConnectionTimeout: 3
+  # udpConnectionRetries Number of retries for SNMP operations
```
omrozowicz-splunk, line 413: let's keep number lowercase

splunk_connect_for_snmp/discovery/loader.py
```diff
@@ -20,9 +23,18 @@
     "DISCOVERY_CONFIG_PATH", "/app/discovery/discovery-config.yaml"
 )
 CHAIN_OF_TASKS_EXPIRY_TIME = os.getenv("CHAIN_OF_TASKS_EXPIRY_TIME", "60")
+LOG_LEVEL = os.getenv("LOG_LEVEL", "DEBUG")
```
omrozowicz-splunk, line 26: Maybe let's have INFO as a default logging level?

splunk_connect_for_snmp/discovery/loader.py
```diff
@@ -20,9 +23,18 @@
...
+
+formatter = CustomisedJSONFormatter()
 
+logger = logging.getLogger(__name__)
```
omrozowicz-splunk, line 30: isn't it a part of a celery task? we could use celery logger instead and avoid problems with inconsistent formatting

cwadhwani-splunk: In this case, the loader is responsible for creating the Celery task, which will then be picked up and executed by the discovery worker. That’s why we’ve used this logger here rather than the Celery logger.
(The inventory loader performs a similar task, and it follows the same logging approach, so we’ve referenced that implementation for consistency.)

omrozowicz-splunk (approved):

PR #1267, opened 2025-10-13
fix: delete redis from bitnami and provide own templates

omrozowicz-splunk:
# Description

This PR:
1. Removes bitnami redis dependency
2. Upgrade redis version to the newest
3. Add redis authentication — both plain text and secret

Fixes # (issue)

## Type of change

- [ ] Dependency update
- [ ] Refactor/improvement

## How Has This Been Tested?

Manual tests, integration tests

## Checklist

### DIFF
--- a/charts/splunk-connect-for-snmp/templates/_helpers.tpl
+++ b/charts/splunk-connect-for-snmp/templates/_helpers.tpl
@@ -14,22 +14,6 @@
 {{- end }}
 {{- end }}
 
-{{- define "splunk-connect-for-snmp.celery_url" -}}
-{{- if and ( eq .Values.redis.architecture "replication" ) .Values.redis.sentinel.enabled  }}
-{{- printf "redis://%s-redis:6379/0" .Release.Name }}
-{{- else }}
-{{- printf "redis://%s-redis-master:6379/0" .Release.Name }}
-{{- end }}
-{{- end }}
-
-{{- define "splunk-connect-for-snmp.redis_url" -}}
-{{- if and ( eq .Values.redis.architecture "replication" ) .Values.redis.sentinel.enabled  }}
-{{- printf "redis://%s-redis:6379/1" .Release.Name }}
-{{- else }}
-{{- printf "redis://%s-redis-master:6379/1" .Release.Name }}
-{{- end }}
-{{- end }}
-
 {{/*
 Create a default fully qualified app name.
 We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
--- a/charts/splunk-connect-for-snmp/templates/redis/redis-standalone-statefulset.yaml
+++ b/charts/splunk-connect-for-snmp/templates/redis/redis-standalone-statefulset.yaml
@@ -1,0 +1,159 @@
+{{- if eq .Values.redis.architecture "standalone" }}
+{{- $bitnamiPVC := printf "redis-data-%s-redis-master-0" .Release.Name }}
+{{- $existingPVC := lookup "v1" "PersistentVolumeClaim" .Release.Namespace $bitnamiPVC }}
+apiVersion: apps/v1
+kind: StatefulSet
+metadata:
+  name: {{ .Release.Name }}-redis
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app: {{ .Release.Name }}-redis
+spec:
+  serviceName: {{ .Release.Name }}-redis
+  replicas: 1
+  selector:
+    matchLabels:
+      app: {{ .Release.Name }}-redis
+  template:
+    metadata:
+      labels:
+        app: {{ .Release.Name }}-redis
+      annotations:
+        checksum/config: {{ include (print $.Template.BasePath "/redis/redis-config.yaml") . | sha256sum }}
+    spec:
+      {{- with .Values.redis.podSecurityContext }}
+      securityContext:
+        runAsUser: {{ .runAsUser }}
+        fsGroup: {{ .fsGroup }}
+      {{- end }}
+      initContainers:
+      - name: fix-permissions
+        image: {{ .Values.redis.image.repository }}:{{ .Values.redis.image.tag }}
+        imagePullPolicy: {{ .Values.redis.image.pullPolicy }}
+        command:
+        - sh
+        - -c
+        - |
+          echo "=== Redis Init: Fixing Permissions ==="
+          echo "Current ownership:"
+          ls -ln /data
+          echo ""
+          echo "Fixing ownership to {{ .Values.redis.podSecurityContext.runAsUser }}:{{ .Values.redis.podSecurityContext.fsGroup }}..."
+          chown -R {{ .Values.redis.podSecurityContext.runAsUser }}:{{ .Values.redis.podSecurityContext.fsGroup }} /data
+          chmod -R 755 /data
+          echo ""
+          echo "New ownership:"
+          ls -ln /data
+          echo "=== Permissions Fixed ==="
+        volumeMounts:
+        - name: redis-data
+          mountPath: /data
+        securityContext:
+          runAsUser: 0  # Must run as root to chown
+      containers:
+      - name: redis
+        image: {{ .Values.redis.image.repository }}:{{ .Values.redis.image.tag }}
+        imagePullPolicy: {{ .Values.redis.image.pullPolicy }}
+        ports:
+        - containerPort: 6379
+          name: redis
+        command:
+        - sh
+        - -c
+        args:
+        - |
+          # Copy config to writable location
+          cp /etc/redis/redis.conf /tmp/redis.conf
+
+          {{- if .Values.redis.auth.enabled }}
+          # Append password at runtime
+          echo "requirepass $REDIS_PASSWORD" >> /tmp/redis.conf
+          {{- end }}
+
+          # Start Redis
+          exec redis-server /tmp/redis.conf
+        {{- if .Values.redis.auth.enabled }}
+        env:
+          - name: REDIS_PASSWORD
+            valueFrom:
+              secretKeyRef:
+                {{- if .Values.redis.auth.existingSecret }}
+                name: {{ .Values.redis.auth.existingSecret }}
+                key: {{ .Values.redis.auth.existingSecretPasswordKey | default "password" }}
+                {{- else }}
+                name: {{ .Release.Name }}-redis-secret
+                key: password
+                {{- end }}
+        {{- end }}
+        volumeMounts:
+        - name: redis-data
+          mountPath: /data
+        - name: redis-config
+          mountPath: /etc/redis
+        resources:
+          {{- toYaml .Values.redis.resources | nindent 10 }}
+        livenessProbe:
+          exec:
+            command:
+            - sh
+            - -c
+            - |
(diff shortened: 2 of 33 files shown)


charts/splunk-connect-for-snmp/templates/_helpers.tpl
```diff
@@ -15,18 +15,36 @@
...
+{{- if .Values.redis.auth.enabled -}}
+  {{- $secretName := .Values.redis.auth.existingSecret | default (printf "%s-redis-secret" .Release.Name) -}}
+  {{- $secret := lookup "v1" "Secret" .Release.Namespace $secretName -}}
+  {{- if $secret -}}
+    {{- $passwordKey := .Values.redis.auth.existingSecretPasswordKey | default "password" -}}
+    {{- $password = index $secret.data $passwordKey | b64dec -}}
+  {{- end -}}
+{{- end -}}
+{{- if $password }}
+{{- printf "redis://:%s@%s-redis:6379/1" $password .Release.Name | quote -}}
```
Kawron: I wonder if this is the most secure solution. We are retrieving the password from a k8s secret, but then "splunk-connect-for-snmp.redis_url" is used as an ENV variable with the password in plain text. This defeats the purpose of having the password secret in the first place.

Kawron (changes requested):

charts/splunk-connect-for-snmp/values.yaml
```diff
@@ -596,9 +596,46 @@ mongodb:
       repository: bitnamilegacy/os-shell
     enabled: true
 redis:
+  # Mode selector: "standalone"
+  # TODO: add "replication"
```
Kawron, line 600: Is this 'todo' left intentionally?

charts/splunk-connect-for-snmp/templates/redis/redis-standalone-statefulset.yaml
```diff
@@ -0,0 +1,147 @@
...
+          ls -ln /data
+          echo ""
+          echo "Fixing ownership to {{ .Values.redis.podSecurityContext.runAsUser }}:{{ .Values.redis.podSecurityContext.fsGroup }}..."
+          chown -R {{ .Values.redis.podSecurityContext.runAsUser }}:{{ .Values.redis.podSecurityContext.fsGroup }} /data
+          chmod -R 755 /data
+          echo ""
+          echo "New ownership:"
+          ls -ln /data
+          echo "=== Permissions Fixed ==="
+        volumeMounts:
```
Kawron, line 48: Will it work if: ".Values.redis.storage.enabled" will be set to false?

charts/splunk-connect-for-snmp/templates/redis/redis-standalone-statefulset.yaml
```diff
@@ -0,0 +1,147 @@
...
+        - name: redis-data
+          mountPath: /data
+        securityContext:
+          runAsUser: 0  # Must run as root to chown
```
Kawron, line 52: Do we need this if we have a fsGroup option? Shouldn't all files be owned by the fsGroup with the secruityContext?

charts/splunk-connect-for-snmp/templates/redis/redis-standalone-statefulset.yaml
```diff
@@ -0,0 +1,147 @@
...
+          # Append password at runtime
+          echo "requirepass $REDIS_PASSWORD" >> /tmp/redis.conf
+          {{- end }}
+
+          # Start Redis
+          exec redis-server /tmp/redis.conf
+        {{- if .Values.redis.auth.enabled }}
+        env:
+        - name: REDIS_PASSWORD
+          {{- if or .Values.redis.auth.existingSecret }}
```
Kawron: "or" with single value

charts/splunk-connect-for-snmp/templates/redis/redis-standalone-statefulset.yaml
```diff
@@ -0,0 +1,147 @@
...
+        - sh
+        - -c
+        args:
+        - |
+          # Copy config to writable location
+          cp /etc/redis/redis.conf /tmp/redis.conf
+
+          {{- if .Values.redis.auth.enabled }}
+          # Append password at runtime
+          echo "requirepass $REDIS_PASSWORD" >> /tmp/redis.conf
```
Kawron, line 70: It is already defined in the config map line 16

charts/splunk-connect-for-snmp/templates/redis/redis-standalone-statefulset.yaml
```diff
@@ -0,0 +1,147 @@
...
+        ports:
+        - containerPort: 6379
+          name: redis
+        command:
+        - sh
+        - -c
+        args:
+        - |
+          # Copy config to writable location
+          cp /etc/redis/redis.conf /tmp/redis.conf
```
Kawron, line 66: Do we need to do this? We shouldn't have to save the password as it is already done in the config map.

And if someone will change the config map, the /etc/redis/redis.conf file will be updated but not the /tmp/redis.conf so for changes to apply the redis container has to be restarted.

So maybe we should stick to the original file, or add to the documentation that the redis has to be restared each time someone changes the config map?

charts/splunk-connect-for-snmp/templates/_helpers.tpl
omrozowicz-splunk: yes, you're right! the fields in _helper are hidden, but then we pass them as env variables that can be easily checked 🙈

charts/splunk-connect-for-snmp/values.yaml
omrozowicz-splunk: the idea is right, replication is left to implement — maybe I should delete the `TODO` keyword though, it shows warning during the commits

charts/splunk-connect-for-snmp/templates/redis/redis-standalone-statefulset.yaml
omrozowicz-splunk: This init job is here in order to make possible migration from bitnami. So the idea is that we already have a redis `pvc` created by bitnami redis (this is dumped in a file in `/data`), but in order for the new redis to read it we have to change permission to whatever is configured in `podSecurityContext`. If we don't do it, the error `read-only` shows. It has to be run as user `0` to do `chown`

charts/splunk-connect-for-snmp/templates/redis/redis-standalone-statefulset.yaml
omrozowicz-splunk: fixed

charts/splunk-connect-for-snmp/templates/redis/redis-standalone-statefulset.yaml
omrozowicz-splunk: deleted from the configmap

charts/splunk-connect-for-snmp/templates/redis/redis-standalone-statefulset.yaml
omrozowicz-splunk: I think in every case when we use secrets this is the scenario that you have to recreate pods in order for them to mount a new secret (same with configmap). I think it is a good idea to add it to the documentation.

mattk42: The restart requirement can be avoided if in the statefulset's pod tempate we add an annotation that includes a checksum of the configmap's values with something like below:

`checksum/config: {{ include (print $.Template.BasePath "templates/redis/redis-config.yaml") . | sha256sum }} `

omrozowicz-splunk: hello @mattk42, thanks for the comment! 😄 I updated the template and tested — it works fine. 
But in this case it'd be more beneficial to have such annotation for secrets. The only thing is that then we'd need to include such annotation on almost every pod (as all of them connect to the redis and have redis secret being mounted to it in order to calculate the connection string). Does it make sense? Is this how it should be done? The alternative way is to document that user has to recreate pods after changing secret's value.

Also, it would be difficult to calculate it as we can either have secret created from the `/redis-redis-secret.yaml` or the secret passed by the `redis.auth.existingSecret`

mattk42: Yeah, that would be challenging and would add to the complexity of the chart as a whole. Personally I do really like automating the restarts, but it may not be worth the complexity in this case especially if it does not happen often.

ui_tests/config/ui_values.yaml
```diff
@@ -3,12 +3,12 @@ UI:
   frontEnd:
     NodePort: 30001
     repository: ghcr.io/splunk/sc4snmp-ui/frontend/container
-    tag: "develop"
+    tag: "1.1.1"
```
ajasnosz, line 6: do we deploy also under latest tag in the ui repository, wouldn't it be better to use it?

charts/splunk-connect-for-snmp/values.yaml
```diff
@@ -23,12 +23,12 @@ UI:
   frontEnd:
     NodePort: 30001
     repository: ghcr.io/splunk/sc4snmp-ui/frontend/container
-    tag: "main"
+    tag: "1.1.1"
```
ajasnosz, line 26: don't we want to revert to main or latest tag?

integration_tests/prepare_splunk.sh
```diff
@@ -17,7 +17,7 @@ create_splunk_hec() {
 
 change_min_free_space() {
   DOCKER_ID=$(sudo docker ps | grep 'splunk/splunk:latest' | awk '{ print $1 }')
-  sudo docker exec --user splunk "$DOCKER_ID" bash -c "echo -e '\n[diskUsage]\nminFreeSpace = 2000' >> /opt/splunk/etc/system/local/server.conf"
+  sudo docker exec --user splunk "$DOCKER_ID" bash -c "echo -e '\n[diskUsage]\nminFreeSpace = 500' >> /opt/splunk/etc/system/local/server.conf"
```
ajasnosz, line 20: Is the 500 our max with the current UI setup?

ajasnosz (changes requested):

charts/splunk-connect-for-snmp/values.yaml
omrozowicz-splunk: imo it is better to use proper versioning to have full ocntrol over user's upgrades. When we keep it as `main` or `latest`, even users using old SC4SNMP versions might have some problems as UI version they use is no longer compatible.

integration_tests/prepare_splunk.sh
omrozowicz-splunk: didn't work with 2000 for sure, maybe it woul work with 1000 (didn't test) but anyway it is minFreeSpace left on a device, so the real free space should be >= than this

ui_tests/config/ui_values.yaml
omrozowicz-splunk: Same as main values.yaml, I think we should avoid using general tags that can frequently change. Every time we want to run those tests with the new UI version we can edit it manually.

ajasnosz (approved):

PR #1284, opened 2025-11-17
fix: add HA templates and modify the code

omrozowicz-splunk:
# Description

This PR adds HA Sentinel option to Redis component. Some of the main changes:
1. Initialise all redis env variables in `construct-redis-url.sh`
2. Add `replication` option to `redis.mode`
3. Add readiness and liveness probe
4. Add annotations to retrigger all pods whenever secrets change

Fixes # (issue)

## Type of change

- [ ] New feature
- [ ] Refactor/improvement

## How Has This Been Tested?

## Checklist

### DIFF
--- a/charts/splunk-connect-for-snmp/templates/_helpers.tpl
+++ b/charts/splunk-connect-for-snmp/templates/_helpers.tpl
@@ -95,3 +95,62 @@ Whether enable polling
 {{- printf "false" }}
 {{- end -}}
 {{- end }}
+
+{{- /*
+Generate Redis environment variables for application pods
+*/ -}}
+{{- define "splunk-connect-for-snmp.redis-env" -}}
+{{- if eq .Values.redis.architecture "replication" }}
+- name: REDIS_MODE
+  value: "replication"
+- name: REDIS_SENTINEL_SERVICE
+  value: {{ .Release.Name }}-redis-sentinel
+- name: REDIS_HEADLESS_SERVICE
+  value: {{ .Release.Name }}-redis-headless
+- name: NAMESPACE
+  value: {{ .Release.Namespace }}
+- name: REDIS_SENTINEL_REPLICAS
+  value: {{ .Values.redis.sentinel.replicas | quote }}
+- name: REDIS_SENTINEL_PORT
+  value: "26379"
+- name: REDIS_MASTER_NAME
+  value: mymaster
+{{- else -}}
+- name: REDIS_MODE
+  value: "standalone"
+- name: REDIS_HOST
+  value: {{ .Release.Name }}-redis
+- name: REDIS_PORT
+  value: "6379"
+{{- end }}
+- name: REDIS_DB
+  value: "1"
+- name: CELERY_DB
+  value: "0"
+{{- if .Values.redis.auth.enabled }}
+- name: REDIS_PASSWORD
+  valueFrom:
+    secretKeyRef:
+      {{- if .Values.redis.auth.existingSecret }}
+      name: {{ .Values.redis.auth.existingSecret }}
+      key: {{ .Values.redis.auth.existingSecretPasswordKey | default "password" }}
+      {{- else }}
+      name: {{ .Release.Name }}-redis-secret
+      key: password
+      {{- end }}
+{{- end -}}
+{{- end }}
+
+{{- /*
+Generate Redis environment variables for application pods
+*/ -}}
+{{- define "splunk-connect-for-snmp.redis-annotations" -}}
+{{- if eq .Values.redis.architecture "replication" }}
+checksum/redis-config: {{ include (print $.Template.BasePath "/redis/redis-ha-config.yaml") . | sha256sum }}
+{{- else -}}
+checksum/redis-config: {{ include (print $.Template.BasePath "/redis/redis-config.yaml") . | sha256sum }}
+{{- end }}
+{{- if .Values.redis.auth.enabled }}
+checksum/redis-secret: {{ include (print $.Template.BasePath "/redis/redis-secret.yaml") . | sha256sum }}
+{{- end }}
+{{- end }}
--- a/charts/splunk-connect-for-snmp/templates/_helpers.tpl
+++ b/charts/splunk-connect-for-snmp/templates/_helpers.tpl
@@ -100,7 +100,7 @@ Whether enable polling
 Generate Redis environment variables for application pods
 */ -}}
 {{- define "splunk-connect-for-snmp.redis-env" -}}
-{{- if eq .Values.redis.architecture "replication" }}
+{{- if eq .Values.redis.architecture "replication" -}}
 - name: REDIS_MODE
   value: "replication"
 - name: REDIS_SENTINEL_SERVICE
@@ -139,18 +139,18 @@ Generate Redis environment variables for application pods
       key: password
       {{- end }}
 {{- end -}}
-{{- end }}
+{{- end -}}
 
 {{- /*
 Generate Redis environment variables for application pods
 */ -}}
 {{- define "splunk-connect-for-snmp.redis-annotations" -}}
-{{- if eq .Values.redis.architecture "replication" }}
+{{- if eq .Values.redis.architecture "replication" -}}
 checksum/redis-config: {{ include (print $.Template.BasePath "/redis/redis-ha-config.yaml") . | sha256sum }}
 {{- else -}}
 checksum/redis-config: {{ include (print $.Template.BasePath "/redis/redis-config.yaml") . | sha256sum }}
-{{- end }}
+{{- end -}}
 {{- if .Values.redis.auth.enabled }}
 checksum/redis-secret: {{ include (print $.Template.BasePath "/redis/redis-secret.yaml") . | sha256sum }}
-{{- end }}
-{{- end }}
+{{- end -}}
+{{- end -}}
--- a/charts/splunk-connect-for-snmp/templates/redis/redis-ha-statefulset.yaml
+++ b/charts/splunk-connect-for-snmp/templates/redis/redis-ha-statefulset.yaml
@@ -1,16 +1,18 @@
-{{- if eq .Values.redis.architecture "standalone" }}
-{{- $bitnamiPVC := printf "redis-data-%s-redis-master-0" .Release.Name }}
-{{- $existingPVC := lookup "v1" "PersistentVolumeClaim" .Release.Namespace $bitnamiPVC }}
+{{- if eq .Values.redis.architecture "replication" }}
 apiVersion: apps/v1
 kind: StatefulSet
 metadata:
-  name: {{ .Release.Name }}-redis-standalone
+  name: {{ .Release.Name }}-redis-ha
   namespace: {{ .Release.Namespace }}
   labels:
     app: {{ .Release.Name }}-redis
+    app.kubernetes.io/name: redis
+    app.kubernetes.io/instance: {{ .Release.Name }}
+    app.kubernetes.io/component: database
 spec:
-  serviceName: {{ .Release.Name }}-redis
-  replicas: 1
+  serviceName: {{ .Release.Name }}-redis-headless
+  replicas: {{ .Values.redis.replicas | default 3 }}
+  podManagementPolicy: Parallel
   selector:
     matchLabels:
       app: {{ .Release.Name }}-redis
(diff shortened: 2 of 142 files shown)


construct-redis-url.sh
```diff
@@ -0,0 +1,65 @@
...
+      REDIS_HA_CHECK="redis://${REDIS_HEADLESS_SERVICE}:${REDIS_PORT}"
+
+    fi
+
+    # Celery broker uses sentinel://
+    : "${CELERY_BROKER_URL:=${SENTINEL_SCHEME}/${CELERY_DB}#master_name=${REDIS_MASTER_NAME}}"
+
+    # RedBeat uses redis-sentinel:// with master_name query
+    : "${REDIS_URL:=${REDBEAT_SCHEME}/${REDIS_DB}#master_name=${REDIS_MASTER_NAME}}"
+    SENTINEL_CHECK="redis://${REDIS_SENTINEL_SERVICE}:${REDIS_SENTINEL_PORT}"
```
Kawron, line 37: shouldn't it be also in the if statment above in case REDIS_PASSWORD is set?

charts/splunk-connect-for-snmp/templates/redis/redis-ha-statefulset.yaml
```diff
@@ -0,0 +1,168 @@
...
+        {{- end }}
+        {{- if .Values.redis.readinessProbe.enabled | default true }}
+        readinessProbe:
+          exec:
+            command:
+            - sh
+            - -c
+            - |
+              {{- if .Values.redis.auth.enabled }}
+              redis-cli -a "$REDIS_PASSWORD" ping
```
Kawron, line 138: This is an overkill, but you can use set REDISCLI_AUTH env to authenticate without passing the password in the command, so it won't be visible in history or process list.  See: https://stackoverflow.com/a/64372000

For this case I think this is okay, because the password is also visible in the config files, but I've just wanted to share this nice feautre : )

charts/splunk-connect-for-snmp/templates/redis/sentinel/redis-sentinel-config.yaml
```diff
@@ -0,0 +1,22 @@
...
+  sentinel.conf: |
+    port 26379
+    dir /data
+    sentinel resolve-hostnames yes
+    sentinel announce-hostnames yes
+    sentinel monitor mymaster {{ .Release.Name }}-redis-ha-0.{{ .Release.Name }}-redis-headless 6379 {{ .Values.redis.sentinel.quorum }}
+    {{- if .Values.redis.auth.enabled }}
+    sentinel auth-pass mymaster ${REDIS_PASSWORD}
+    {{- end }}
+    sentinel down-after-milliseconds mymaster 5000
```
Kawron: Maybe we could parametrize this parameters? Would be there any benefit from fine-tuning those values for customer? If not, I would leave it as it

charts/splunk-connect-for-snmp/templates/redis/sentinel/redis-sentinel-statefulset.yaml
```diff
@@ -0,0 +1,84 @@
...
+kind: StatefulSet
+metadata:
+  name: {{ .Release.Name }}-redis-sentinel
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app: {{ .Release.Name }}-redis-sentinel
+    app.kubernetes.io/component: sentinel
+spec:
+  serviceName: {{ .Release.Name }}-redis-sentinel
+  replicas: {{ .Values.redis.sentinel.replicas | default 3 }}
```
Kawron, line 12: I wonder if we need "default" here? It is already set to 3 in values.yaml

charts/splunk-connect-for-snmp/templates/ui/role-binding.yaml
```diff
@@ -3,11 +3,11 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: RoleBinding
 metadata:
   name: job-robot
-  namespace: sc4snmp
+  namespace: {{ .Release.Namespace | quote }}
```
Kawron, line 6: when searching for: "namespace: sc4snmp" in IDE I still found some occurances. Shouldn't it be cleaned as well?

charts/splunk-connect-for-snmp/templates/worker/_helpers.tpl
```diff
@@ -91,16 +91,9 @@ Common labels
...
-  value: {{ .Release.Name }}-redis
-- name: REDIS_PORT
-  value: "6379"
-- name: REDIS_DB
-  value: "1"
-- name: CELERY_DB
-  value: "0"
 - name: SC4SNMP_VERSION
   value: {{ .Chart.Version | default "0.0.0" }}
+{{- include "splunk-connect-for-snmp.redis-env" . | nindent 0 }}
```
Kawron: I think that nident with 0 does not do anything

charts/splunk-connect-for-snmp/templates/redis/sentinel/redis-sentinel-statefulset.yaml
```diff
@@ -0,0 +1,84 @@
...
+        - name: sentinel-data
+          mountPath: /data
+        resources:
+          {{- toYaml .Values.redis.sentinel.resources | nindent 10 }}
+        livenessProbe:
+          exec:
+            command:
+            - sh
+            - -c
+            - redis-cli -p 26379 ping
```
Kawron, line 57: Will it work if redis is password protected? Password is not passed by -a or REDISCLI_AUTH env

charts/splunk-connect-for-snmp/templates/redis/sentinel/redis-sentinel-statefulset.yaml
```diff
@@ -0,0 +1,84 @@
...
+          initialDelaySeconds: 30
+          periodSeconds: 10
+        readinessProbe:
+          exec:
+            command:
+            - sh
+            - -c
+            - redis-cli -p 26379 sentinel get-master-addr-by-name mymaster | grep snmp-redis-ha-0
```
Kawron: same as above

Kawron: also here we hardcoded release name as snmp whereas in otherplaces it is templated {{ .Release.Name }}

Kawron (changes requested):

charts/splunk-connect-for-snmp/templates/redis/redis-ha-statefulset.yaml
omrozowicz-splunk: I use env to authenticate anyway, this is being rendered to `redis-cli -a "$REDIS_PASSWORD" ping`, `$REDIS_PASSWORD` is not becoming visible from the k8s perspective

construct-redis-url.sh
omrozowicz-splunk: this is what I thought too, but when I tested that it didn't work at all with password 😅 I can check it once again just to be sure

charts/splunk-connect-for-snmp/templates/redis/sentinel/redis-sentinel-statefulset.yaml
omrozowicz-splunk: this is what we generally do in this project as a fallback, not sure if it is necessary, but I know sometimes users edit our values.yaml directly in the local package instead of creating their own 🙈

charts/splunk-connect-for-snmp/templates/ui/role-binding.yaml
omrozowicz-splunk: Yes, this is an issue I found while looking at those UI files. I think I deleted it from all the templates but didn't run `make render` after that so you probably see it in rendered manifests. Let me update it.

charts/splunk-connect-for-snmp/templates/worker/_helpers.tpl
omrozowicz-splunk: `nindent 0` adds a newline before the content and indents all lines by 0 spaces. In this case, alternatively I could use it like:
```
{{ include "splunk-connect-for-snmp.redis-env" .  }}
```

construct-redis-url.sh
omrozowicz-splunk: ok, I know why — `sed -i "s/\${REDIS_PASSWORD}/$REDIS_PASSWORD/g" /data/sentinel.conf` I have in config is handling Redis password (to access redis instances and know their roles) and not the sentinel's one. Sentinel instances are running without password, and from what I read it is standard practice for internal setups. It's enough to have redis passwrod protected. I'll test again if we need password in `SENTINEL_SCHEME` and `REDBEAT_SCHEME`

charts/splunk-connect-for-snmp/templates/redis/sentinel/redis-sentinel-config.yaml
omrozowicz-splunk: not sure about the parameters, but for sure I can use that env variable for `mymaster`

charts/splunk-connect-for-snmp/templates/redis/sentinel/redis-sentinel-statefulset.yaml
omrozowicz-splunk: check: https://github.com/splunk/splunk-connect-for-snmp/pull/1284#discussion_r2541102185 -> sentinel does not have own authentication

charts/splunk-connect-for-snmp/templates/redis/sentinel/redis-sentinel-statefulset.yaml
omrozowicz-splunk: I deleted it completely as after first failover it doesn't work (as other replica is a master)

charts/splunk-connect-for-snmp/templates/ui/role-binding.yaml
omrozowicz-splunk: done

Kawron (approved):

charts/splunk-connect-for-snmp/templates/_helpers.tpl
```diff
@@ -94,4 +94,63 @@ Whether enable polling
...
+- name: REDIS_HEADLESS_SERVICE
+  value: {{ .Release.Name }}-redis-headless
+- name: NAMESPACE
+  value: {{ .Release.Namespace }}
+- name: REDIS_SENTINEL_REPLICAS
+  value: {{ .Values.redis.sentinel.replicas | quote }}
+- name: REDIS_SENTINEL_PORT
+  value: "26379"
+- name: REDIS_MASTER_NAME
+  value: mymaster
```
ajasnosz, line 117: can we set different default name?

docs/microk8s/configuration/redis-configuration.md
```diff
@@ -55,30 +59,156 @@ redis:
...
+    storageClassName: nfs
+    accessModes:
+      - ReadWriteMany
+    size: 3Gi
+```
+
+!!!note
+    The `storageClassName` must point to a `StorageClass` that supports network-attached volumes with `ReadWriteMany` (or `ReadWriteOnce` with cross-node support). Examples: NFS, Ceph, EBS, GCP Persistent Disk, Azure Disk.
+
+For MicrokK8s on Ubuntu, you can enable NFS this way:
```
ajasnosz: typo in microk8s

docs/microk8s/configuration/redis-configuration.md
```diff
@@ -55,30 +59,156 @@ redis:
...
+
+1. Enable NFS client on ALL nodes:
```
ajasnosz, line 173: maybe we want to emphasize more that this is just an example, not solution we recommend and support while configuring?

charts/splunk-connect-for-snmp/templates/_helpers.tpl
omrozowicz-splunk: Any name would work, the only thing is that it must be the same everywhere. From what I learned it is better not to give user option to change this name, as it is easy to break this way.

docs/microk8s/configuration/redis-configuration.md
omrozowicz-splunk: fixed

docs/microk8s/configuration/redis-configuration.md
omrozowicz-splunk: I added `EXAMPLE:` before the section

ajasnosz (approved):

PR #1287, opened 2025-11-27
chore: migrate auto-discovery feature to make compatible with lextuio's pysnmp and python 3.13

gmaiyani-crest:
# Description

- Migrate auto-discovery feature to make compitable with lextuio's pysnmp and python 3.13
- Replace the thread-pool executor with semaphore asyncio for concurrency control

Fixes # (issue)

## Type of change

- [x] Dependency update
- [x] Refactor/improvement

## Checklist


splunk_connect_for_snmp/common/discovery_record.py
```diff
@@ -0,0 +1,105 @@
...
+
+from splunk_connect_for_snmp.common.hummanbool import human_bool
+
+DiscoveryStr = Union[None, str]
+DiscoveryInt = Union[None, int]
+DiscoveryBool = Union[None, bool]
+DiscoveryList = Union[None, List[dict]]
+
+
+class DiscoveryRecord(BaseModel):
```
omrozowicz-splunk: DiscoveryRecord is almost the same as InventoryRecord, right? Can we create some upper level class that both Inventory Record and Discovery record will derive from and put there all the common fields like:
```
    @validator("port", pre=True)
    def port_validator(cls, value):
        if value is None:
            return 161
        if isinstance(value, int) and value >= 1 or value <= 65535:
            return value
        else:
            raise ValueError("field port must be an integer between 1 and 65535")
```
so then when we want to change it it is the only point of control

splunk_connect_for_snmp/snmp/auth.py
```diff
@@ -82,29 +87,33 @@ async def get_security_engine_id(logger, ir: InventoryRecord, snmp_engine: SnmpE
...
+        observer_context, error_indication, rt.address
     )
-    logger.debug(f"securityEngineId={security_engine_id} for device {ir.address}")
+    logger.debug(f"securityEngineId={security_engine_id} for device {rt.address}")
     return security_engine_id
 
 
-async def setup_transport_target(ir: InventoryRecord):
-    ip = get_ip_from_socket(ir) if IPv6_ENABLED else ir.address
+async def setup_transport_target(rt: RecordType):
```
omrozowicz-splunk: If we derive both:
```
class BaseRecord:
    ...

class DiscoveryRecord(BaseRecord):
    ...

class InventoryRecord(BaseRecord):
    ...
```
then I think we can use `BaseRecord` instead of creating unions

ajasnosz (approved):

splunk_connect_for_snmp/snmp/auth.py
gmaiyani-crest: sure, I have made the changes according to your suggestions.

splunk_connect_for_snmp/common/discovery_record.py
gmaiyani-crest: Yes it is, I have made the changes according to your suggestions.

omrozowicz-splunk (approved):

PR #1293, opened 2025-12-08
fix: replace Bitnami MongoDB chart with custom MongoDB templates (standalone + replicaset)

omrozowicz-splunk:
# Description

This PR removes the Bitnami MongoDB chart dependency and replaces it with fully internal MongoDB templates that support both Standalone and ReplicaSet modes.
The redesign improves startup reliability, simplifies configuration, supports PVC migration, and provides full control over authentication handling.

Key Changes:

**New MongoDB StatefulSet Architecture**
Separate templates for:

- standalone mode (single pod)
- replicaset mode (multi‑pod headless)

Clearer structure with no large conditional templates.

**PVC Migration for Standalone**

- If a Bitnami PVC (datadir-<release>-mongodb-0) exists, it is reused automatically.
- Enables seamless migration without data copy.
- ReplicaSet always uses fresh PVCs.

**Authentication Simplified**

Users may:

- Provide credentials directly in values.yaml, or
- Supply an existingUserSecret.
- The chart automatically generates the replicaKey (secure, 64‑char).

**Improved ReplicaSet Initialization Job**

Uses official kubectl image.
Waits for:

- All pods to be Ready
- DNS + network reachability (nc -z)
- Runs rs.initiate() via kubectl exec on pod 0.
- Idempotent: safe to rerun; skips if RS already initialized.

**Application Wait Logic Updated**

- Waits for PRIMARY using no‑auth first (MongoDB topology discovery).
- Then uses credentials once available.
- Avoids failures during early startup when the root user does not exist yet.

**Clean Secret Model**

- One secret for user credentials (user‑managed or chart‑generated).
- One secret for replicaKey (chart‑generated only).
- No Bitnami env vars or secret formats.

**Bitnami Removed Completely**

No Bitnami containers, init scripts, env vars, or PVC names (except migration detection).

Fixes # (issue)

## Type of change

- [ ] Dependency update
- [ ] New feature
- [ ] Refactor/improvement

## How Has This Been Tested?

## Checklist


construct-connection-strings.sh
```diff
@@ -61,8 +65,40 @@ if [ -z "$REDIS_URL" ] || [ -z "$CELERY_BROKER_URL" ]; then
...
+############################
+      ### MongoDB ###
+############################
+
+  # Build MongoDB URI from environment variables
+  if [ -n "$MONGODB_PASSWORD" ]; then
+    # With authentication
+    if [ -n "$MONGODB_REPLICA_SET" ]; then
+      # Replica set
+      export MONGO_URI="mongodb://${MONGODB_USERNAME}:${MONGODB_PASSWORD}@${MONGODB_HOST}/${MONGODB_DATABASE}?replicaSet=${MONGODB_REPLICA_SET}&authSource=${MONGODB_AUTH_SOURCE:-admin}&readPreference=primary"
```
Kawron, line 77: Here we are setting: `readPreference=primary` but on line 85 where we have replica set without authentication we don't have it. Why is that? Is it related to the authentication mechanism?

splunk_connect_for_snmp/common/common.py
```diff
@@ -62,3 +68,55 @@ def convert_to_float(value: typing.Any, ignore_error: bool = False) -> typing.An
...
+
+    mongo_uri = os.getenv("MONGO_URI")
+
+    if not mongo_uri:
+        logger.warning("MONGO_URI not set, exiting application")
+        sys.exit(1)
+
+    logger.info(f"Waiting for MongoDB ReplicaSet to be ready and elect the primary...")
+
+    for attempt in range(1, max_retries + 1):
```
Kawron, line 107: If for some reason the `replicaSet=` wouldn't be in mongo_uri, then we would have a very hard to debug problem as the loop would just end without any logs. So I would move the:
```python
            if attempt >= max_retries:
                logger.info(f"MongoDB not ready after {max_retries * retry_interval}s")
                logger.info(f"   Error: {e}")
                sys.exit(1)
```
after the loop.

Also I think it is unintuitive to use the exception mechanism to handle timeout. I would move the `sleep` part and the `if attempt` part out of the catch block and just handle the ServerSelectionTimeoutError and ConnectionFailure which I guess are for `MongoClient()` class. Sth more or less like this:

```python
    for attempt in range(1, max_retries + 1):
        if attempt % 6 == 0:  # Print every 30 seconds
        logger.info(
            f"  Still waiting... ({attempt}/{max_retries}) - {e.__class__.__name__}"
        )

        if attempt != 1:
            time.sleep(retry_interval)
        try:
            # Try to connect
            client = MongoClient(
                mongo_uri, serverSelectionTimeoutMS=5000, connectTimeoutMS=5000
            )

            # Execute a simple operation to verify PRIMARY exists
            client.admin.command("ping")

            # For replica sets, verify PRIMARY exists
            if "replicaSet=" in mongo_uri:
                if client.primary is None:
                    continue
                logger.info(f"PRIMARY found: {client.primary}")

            client.close()
            logger.info("MongoDB is ready")
            return

        except (ServerSelectionTimeoutError, ConnectionFailure) as e:
            # log sth about those exceptions
           pass
            
logger.info(f"MongoDB not ready after {max_retries * retry_interval}s")
            logger.info(f"   Error: {e}")
            sys.exit(1)
```

docs/microk8s/configuration/redis-configuration.md
```diff
@@ -44,7 +44,7 @@ redis:
   # Storage
   storage:
     enabled: true
-    storageClassName: microk8s-hostpath
+    storageClassName: ""
```
Kawron, line 47: If empty means 'default' isn't it just enough to not provide the storageClassName parameter. Or if we want to explicite use default maybe it would be better to just write `storageClassName: default`?

Kawron (changes requested):

charts/splunk-connect-for-snmp/templates/mongodb/mongodb-init-rs-job.yaml
```diff
@@ -0,0 +1,97 @@
...
+
+          echo "=========================================="
+          echo " MongoDB Replica Set Initialization"
+          echo "=========================================="
+          echo ""
+
+          echo "Waiting for all {{ .Values.mongodb.replicaCount }} pods to be ready..."
+          {{- range $i := until (int .Values.mongodb.replicaCount) }}
+          kubectl wait --for=condition=ready pod/{{ $.Release.Name }}-mongodb-{{ $i }} --timeout={{ $.Values.mongodb.replicaInitJob.timeout | quote }}s
+          echo "  ✅ Pod {{ $i }} ready"
```
Kawron: I'm personally against using emojis in the code. Also, if someone would have some archaic terminal it may not display correctly

Kawron (changes requested):

charts/splunk-connect-for-snmp/templates/mongodb/mongodb-ha-statefulset.yaml
```diff
@@ -0,0 +1,176 @@
...
+      name: data
+    spec:
+      accessModes:
+      - {{ .Values.mongodb.persistence.accessMode | default "ReadWriteOnce" }}
+      {{- if .Values.mongodb.persistence.storageClassName }}
+      storageClassName: {{ .Values.mongodb.persistence.storageClassName }}
+      {{- end }}
+      resources:
+        requests:
+          storage: {{ .Values.mongodb.persistence.size | default "10Gi" }}
```
ajasnosz, line 174: Do we base the default on some research or mongo suggestions?

charts/splunk-connect-for-snmp/values.yaml
```diff
@@ -642,7 +613,7 @@ redis:
   # Storage
   storage:
     enabled: true
-    storageClassName: microk8s-hostpath
+    storageClassName: ""
```
ajasnosz, line 639: what will be used if we don't specify any storage? will it have to be set up manually every time?

docs/microk8s/configuration/mongo-configuration.md
```diff
@@ -1,34 +1,300 @@
...
-  resources: 
-    limits:
-      cpu: 2
-      memory: 2Gi
+  # Mode selector: "standalone", "replication"
+  mode: standalone
+
+  # Replica set configuration (used only when mode = "replication")
+  replicaCount: 3
+  replicaSetName: rs0
```
ajasnosz: If the mode is set to standalone as right now, would this config just not be read when creating the mongodb? or it shouldn't be added to values.yaml then?
If so then it might be confusing for people if we write it like that in docs.

docs/microk8s/configuration/mongo-configuration.md
```diff
@@ -1,34 +1,300 @@
...
+    rootUser: "admin"
+    rootPassword: ""                  # Set if auth.enabled: true
+    existingUserSecret: ""            # Or reference existing secret
+    rootUserKey: "root-user"
+    rootPasswordKey: "root-password"
+
+  # Image
+  image:
+    repository: mongo
+    tag: "7.0"
```
ajasnosz: can we use the same tag as in code - 8.2.2?

docs/microk8s/configuration/mongo-configuration.md
```diff
@@ -1,34 +1,300 @@
...
+* Single-node environments
+* Development and testing
+* Non-critical workloads
+
+Characteristics:
+
+* Resources: 1 MongoDB pod
+* Complexity: Low
+* Recovery time: ~30-60 seconds (Kubernetes reschedules pod on node failure)
+* No automatic failover
```
ajasnosz: isn't this copied from standalone, instead of being written for replica mode?

docs/microk8s/configuration/mongo-configuration.md
```diff
@@ -1,34 +1,300 @@
...
+    rootUser: "admin"
+    rootPassword: "your_secure_password_here"
+```
+
+#### Using Existing Kubernetes Secret
+
+To use an existing Kubernetes Secret, first create it:
+
+```yaml
+kubectl create secret generic prod-mongodb-secret \
```
ajasnosz: shouldn't the correct namespace be attached here?

docs/microk8s/configuration/mongo-configuration.md
```diff
@@ -1,34 +1,300 @@
...
+    enabled: true
+    existingUserSecret: "prod-mongodb-secret-with-different-keys"
+    rootUserKey: "my-username-key"
+    rootPasswordKey: "my-password-key"
+```
+
+
+### Migration from Bitnami MongoDB
+
+The chart automatically detects and migrates data from existing Bitnami MongoDB deployments in standalone mode only:
```
ajasnosz: only in standalone mode:

docs/microk8s/configuration/mongo-configuration.md
```diff
@@ -1,34 +1,300 @@
...
+
+1. Detects Bitnami PVC: datadir-<release>-mongodb-0
+2. Reuses the PVC if found (preserves data)
+3. Init container fixes file permissions for compatibility
+4. If no existing PVC is found, creates a new one
+
+No manual intervention required — simply upgrade your deployment with the new chart.
+
+!!!note
```
ajasnosz: maybe it should be put as !!!warning ?

docs/microk8s/configuration/mongo-configuration.md
```diff
@@ -1,34 +1,300 @@
...
+    Migration between Bitnami MongoDB and the new chart is possible only to standalone mode. For using replication mode, please reinstall SC4SNMP with a fresh MongoDB deployment.
+
+### Replica Set Initialization
+
+When deploying in replication mode, the chart automatically:
+
+1. Deploys a headless service for stable pod DNS
+2. Creates all MongoDB pods with replica set configuration
+3. Runs a Kubernetes Job to initialize the replica set
+4.Waits for PRIMARY election (typically 10-15 seconds)
```
ajasnosz: missing space

docs/microk8s/configuration/mongo-configuration.md
```diff
@@ -1,34 +1,300 @@
...
+The initialization job:
+
+1. Waits for all pods to be ready
+2. Verifies network connectivity between pods
+3. Runs rs.initiate() from inside pod-0
+4. Is idempotent (safe to re-run)
+
+You can monitor initialization progress:
+```bash
+kubectl logs -f job/<release-name>-mongodb-init-rs
```
ajasnosz: shouldn't namespace be added there? are they all created in general?

splunk_connect_for_snmp/common/hummanbool.py
```diff
@@ -62,3 +68,55 @@ def convert_to_float(value: typing.Any, ignore_error: bool = False) -> typing.An
 def disable_mongo_logging():
     logging.getLogger("mongo").setLevel(logging.CRITICAL)
     logging.getLogger("pymongo").setLevel(logging.CRITICAL)
+
+
+def wait_for_mongodb_replicaset(logger, max_retries=120, retry_interval=5):
```
ajasnosz: I think it shold me moved to different file, or the filename should be changed as it doesn't reflect what we have here.

ajasnosz (changes requested):

construct-connection-strings.sh
omrozowicz-splunk: I think I just forgot to put it here, it is good practice to have `readPreference=primary` as this makes sure we read from primary (although reading from secondary would also work, that's why I didn't notice this issue before). Fixed.

splunk_connect_for_snmp/common/common.py
omrozowicz-splunk: That makes sense, actually I modified it even more as I had some problems with loggers. Also decided to log out every attempt as it take some time (around 10s). Please take a look.

docs/microk8s/configuration/redis-configuration.md
omrozowicz-splunk: I think using empty storageClass is a best practice in Helm templates, using `default` would work only if the storageClass named explicitly `default` exists. Also this is the way bitnami mongodb uses: https://github.com/bitnami/charts/blob/8b08cd4774e6847cf3e76f39622d033404de4a4b/bitnami/mongodb/values.yaml#L24

charts/splunk-connect-for-snmp/templates/mongodb/mongodb-init-rs-job.yaml
omrozowicz-splunk: Agree, removed

charts/splunk-connect-for-snmp/values.yaml
omrozowicz-splunk: If no default storage is specified, the default for the cluster is used. This is according to the [pvc k8s documentation:](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#using-a-storageclass)  ` storageClassName: "" # Empty string must be explicitly set otherwise default StorageClass will be se`

omrozowicz-splunk: also I noticed this is what bitnami charts have, it is possible that in case no storageclass is defined it will fail (not sure if that's ever a case though), but it is still better than hardcoding microk8s storage when we want to move from keeping only to one k8s platform IMO

docs/microk8s/configuration/mongo-configuration.md
omrozowicz-splunk: > If the mode is set to standalone as right now, would this config just not be read when creating the mongodb

yes

Do you think it is better to hide it for good?

charts/splunk-connect-for-snmp/templates/mongodb/mongodb-ha-statefulset.yaml
omrozowicz-splunk: It was suggested by ai as storage for middle sized collection, in previous mongodb we used 8Gi (bitnami's default). Alternatively I can change it to 8Gi.

docs/microk8s/configuration/mongo-configuration.md
omrozowicz-splunk: done

docs/microk8s/configuration/mongo-configuration.md
omrozowicz-splunk: right 😅 updated

docs/microk8s/configuration/mongo-configuration.md
omrozowicz-splunk: yes, fixed

docs/microk8s/configuration/mongo-configuration.md
omrozowicz-splunk: updated

docs/microk8s/configuration/mongo-configuration.md
omrozowicz-splunk: fixed

docs/microk8s/configuration/mongo-configuration.md
omrozowicz-splunk: added

splunk_connect_for_snmp/common/hummanbool.py
omrozowicz-splunk: renamed file to `common.py`

charts/splunk-connect-for-snmp/templates/mongodb/mongodb-ha-statefulset.yaml
ajasnosz: I think it's ok to keep 10, just wanted to know the origin of the value.

docs/microk8s/configuration/mongo-configuration.md
ajasnosz: I don't think we need to hide it, it's more about how it is structured in this documentation. Both of the configs are right next to each other but they can't be used together. We are giving the command but if someone just skips with the eyes and copy all of the config from the page then they could have wrong set up, so maybe we could separate it somehow? or change the mode to replication in this example

omrozowicz-splunk: I set default mode to replication

docs/microk8s/configuration/mongo-configuration.md
omrozowicz-splunk: done

Kawron (approved):

ajasnosz (approved):

ajasnosz (approved):

PR #1334, opened 2026-02-24
Fix: inventory error in dc

### LINKED ISSUE #1323: Bug with empty profiles in docker compose
If the SC4SNMP is only configured to use traps, and has no device for polling and removed worker-pollers the inventory still requires to set some profile in the scheduler-config.yaml.

ajasnosz:
# Description

Solution to #1323. 
The issue is with the docker compose when the inventory.csv file is not existing. The docker will then create the file as directory, and our code would raise an exception, and restart the inventory container in a loop. The solution is to catch this specific error when it is created: IsADirectoryError, add the information in logs, and proceed as if the inventory file is empty. 

Fixes # (issue)

## Type of change

- [x] Bug fix

## How Has This Been Tested?

manually, integration

## Checklist

### DIFF
--- a/splunk_connect_for_snmp/common/inventory_processor.py
+++ b/splunk_connect_for_snmp/common/inventory_processor.py
@@ -130,19 +130,19 @@ class InventoryProcessor:
                 )
                 return [], {}
 
-            for inventory_line in ir_reader:
-                self.process_line(inventory_line)
-            for source_record in self.single_hosts:
-                address = source_record["address"]
-                port = source_record.get("port")
-                host = transform_address_to_key(address, port)
-                was_present = self.hosts_from_groups.get(host, None)
-                if was_present is None:
-                    self.inventory_records.append(source_record)
-                else:
-                    logger.warning(
-                        f"Record: {host} has been already configured in group. Skipping..."
-                    )
+        for inventory_line in ir_reader:
+            self.process_line(inventory_line)
+        for source_record in self.single_hosts:
+            address = source_record["address"]
+            port = source_record.get("port")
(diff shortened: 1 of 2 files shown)


splunk_connect_for_snmp/common/inventory_processor.py
```diff
@@ -118,9 +118,16 @@ def get_all_hosts(self):
...
+            try:
+                with open(INVENTORY_PATH, encoding="utf-8") as csv_file:
+                    logger.info(f"Loading inventory from {INVENTORY_PATH}")
+                    ir_reader = list(DictReader(csv_file))
+            except IsADirectoryError:
+                logger.error(
+                    f"Inventory file {INVENTORY_PATH} not correctly created in the deployment.",
+                    exc_info=True,
+                )
+                ir_reader = []
```
omrozowicz-splunk: maybe we can already return empty lists instead of going through the lest of function anyway? 🤔

ajasnosz: updated

omrozowicz-splunk: there are two problems with that:
1. this code will never execute:
```
        if CONFIG_FROM_MONGO:
            logger.info("Loading inventory from inventory_ui collection")
            ir_reader = list(self.inventory_ui_collection.find({}, {"_id": 0}))
```
2. `finally` surpresses and other error that might happen and it won't be logged out at all which makes troubleshooting harder, I think we should avoid using this

omrozowicz-splunk (approved):

PR #1342, opened 2026-03-04
Feat: discover traps engineid

### LINKED ISSUE #1280: SNMPv3 engine ID scalability
**If your issue is not a bug or a feature request, please raise a support ticket through our support portal (Splunk.com > Support > Support Portal). This will help us resolve your issue more efficiently and provide you with better assistance. For more information on how to work with the Splunk Support, please refer to [this guide](https://splunk.my.site.com/customer/s/article/Working-with-Support-and-the-Support-Portal).**

** What is the SC4SNMP version? 1.12.1

**Is your feature request related to a problem? Please describe.**
We require SC4SNMP to support receiving traps of thousands of different SNMP v3 engine IDs. As per RFC 3411, each engine ID must be unique.

RFC 3411      Architecture for SNMP Management Frameworks  December 2002

3.1.1.1 snmpEngineID

   Within an administrative domain, an snmpEngineID is the unique and unambiguous identifier of an SNMP engine.  Since there is a one-to-one association between SNMP engines and SNMP entities, it also uniquely and unambiguously identifies the SNMP entity within that administrative domain.

**Describe the solution you'd like**
- SNMP_V3_SECURITY_ENGINE_ID to accept a wildcard (*) to allow traps from any engine ID, subject to format-compliant checks (Pages 40-41 of https://www.ietf.org/rfc/rfc3411.txt)
or
- Develop a scalable method to incorporate thousands of allowed engine IDs within the trap configuration

**Describe alternatives you've considered**
Use v2 traps, but it's not an option due to security flaws

**Additional context**

ajasnosz:
# Description

Dynamic discovery of security engine id. If enabled the security engine id will be extracted from the trap message, added to know engine id list and then the traps would be accepted and parsed.
This PR also adds keeping engine id in the mogno db collection. 

Fixes # (issue)

## Type of change

- [x] New feature

## How Has This Been Tested?

manual

## Checklist


splunk_connect_for_snmp/common/collection_manager.py
```diff
@@ -217,3 +219,44 @@ def merge_profiles_from_ui(self, active_profiles):
             key = list(pr.keys())[0]
             profile = pr[key]
             self.assign_profiles_to_dict(active_profiles, key, profile)
+
+
+class EngineIdManager:
```
omrozowicz-splunk, line 224: why can't we derive here from `CollectionManager` like in case of other collections? 🤔

splunk_connect_for_snmp/common/collection_manager.py
```diff
@@ -217,3 +219,44 @@ def merge_profiles_from_ui(self, active_profiles):
...
+    COLLECTION_NAME = "engine_id_records"
+
+    def __init__(self, mongo):
+        self.collection = mongo.sc4snmp[self.COLLECTION_NAME]
+        self._ensure_indexes()
+        self._cached_engine_ids = set(self.collection.distinct("security_engine_id"))
+
+    def _ensure_indexes(self):
```
omrozowicz-splunk, line 232: maybe `schema_migration.py` is a better place to do that?

splunk_connect_for_snmp/common/collection_manager.py
```diff
@@ -217,3 +219,44 @@ def merge_profiles_from_ui(self, active_profiles):
...
+            {"security_engine_id": security_engine_id}, {"_id": 1}
+        )
+        if existing is None:
+            self.collection.insert_one(record)
+        self._cached_engine_ids.add(security_engine_id)
+
+    def get_engine_ids_for_host(self, host):
+        return list(self.collection.find({"host": host}, {"_id": 0}))
+
+    def get_all_engine_ids(self):
```
omrozowicz-splunk, line 258: if you derive from `CollectionManager` you can just use `return_collection_once` instead

splunk_connect_for_snmp/common/collection_manager.py
omrozowicz-splunk: ok I see it is there as well

rendered/manifests/tests/splunk-connect-for-snmp/templates/common/traps-config.yaml
```diff
@@ -16,3 +16,6 @@ data:
         2c:
         - public
         - homelab
+
```
omrozowicz-splunk, line 19: why is it always here by default? we need to have something?

ajasnosz: yes, we always had the default security engine id as "80003a8c04"

splunk_connect_for_snmp/common/collection_manager.py
ajasnosz: We can, but the behaviour we expect from this collection is different. We load profiles/devices from config in one big bulk and replace the old one, while traps engine ids are added one by one when they are discovered during application runtime, also the engine ids should be kept in mongodb and not replaced/removed from mongodb until we explicitly say to do so. So the method that would be inherited won't be used in this case.

omrozowicz-splunk: okay

splunk_connect_for_snmp/common/collection_manager.py
omrozowicz-splunk: probably doesn't make sense because of -> https://github.com/splunk/splunk-connect-for-snmp/pull/1342#discussion_r2905102125

omrozowicz-splunk (approved):

PR #1343, opened 2026-03-05
chore: restructure integration tests and add local test runners

pkalevar16:
…t-tests]

# Description

Fixes # (issue)

## Type of change

- [ ] Dependency update
- [ ] Bug fix
- [ ] New feature
- [ ] Refactor/improvement

## How Has This Been Tested?

## Checklist


integration_tests/scripts/automatic_setup_compose.sh
```diff
@@ -93,24 +107,35 @@ sudo docker run -d -p 8000:8000 -p 8088:8088 -p 8089:8089 -e SPLUNK_GENERAL_TERM
...
 echo $(green "Setting up docker compose configuration")
 cp ../docker_compose/* .
+
 # Define the filenames for the variables
-SCHEDULER_CONFIG_FILE="scheduler-config.yaml"
-TRAPS_CONFIG_FILE="traps-config.yaml"
-INVENTORY_FILE="inventory-tests.csv"
+#SCHEDULER_CONFIG_FILE="scheduler-config.yaml"
+#TRAPS_CONFIG_FILE="traps-config.yaml"
+#INVENTORY_FILE="inventory-tests.csv"
```
ajasnosz: if the commented lines are no longer needed or duplicated please remove them

integration_tests/scripts/automatic_setup_compose.sh
```diff
@@ -200,7 +225,8 @@ awk -v scheduler_path="$SCHEDULER_CONFIG_FILE_ABSOLUTE_PATH" \
 mv "$TEMP_ENV_FILE" .env
 
 
-sed -i "s/###LOAD_BALANCER_ID###/$(hostname -I | cut -d " " -f1)/" inventory-tests.csv
+#sed -i "s/###LOAD_BALANCER_ID###/$(hostname -I | cut -d " " -f1)/" inventory-tests.csv
```
ajasnosz: same here

integration_tests/scripts/automatic_setup_microk8s.sh
```diff
@@ -95,14 +107,25 @@ sudo docker run -d -p 8000:8000 -p 8088:8088 -p 8089:8089 -e SPLUNK_GENERAL_TERM
...
+"$SCRIPT_DIR/prepare_splunk.sh"
+"$SCRIPT_DIR/install_sck.sh"
+# chmod u+x prepare_splunk.sh
+# echo $(green "Preparing Splunk instance")
+# ./prepare_splunk.sh
+# ./install_sck.sh
+
+# sed -i "s/###SPLUNK_TOKEN###/$(cat hec_token)/" values.yaml
+# sed -i "s/###LOAD_BALANCER_ID###/$(hostname -I | cut -d " " -f1)/" values.yaml
+
```
ajasnosz: same here

integration_tests/scripts/automatic_setup_microk8s.sh
```diff
@@ -120,11 +143,16 @@ sudo systemctl enable iscsid
 yes $(hostname -I | cut -d " " -f1)/32 | sudo microk8s enable metallb
 sudo microk8s status --wait-ready
 
-cd ../charts/splunk-connect-for-snmp
+# cd ../charts/splunk-connect-for-snmp
+# sudo microk8s helm3 dep update
+# cd ../../integration_tests
+
```
ajasnosz: same here

integration_tests/scripts/run_local_docker_tests.sh
```diff
@@ -0,0 +1,471 @@
...
+
+
+# ===== INSTALL DOCKER + DOCKER COMPOSE =====
+install_docker() {
+
+if ! command -v docker >/dev/null 2>&1; then
+    step "Installing Docker"
+
+    sudo apt-get update
+    sudo apt-get install -y docker.io curl
```
ajasnosz: let's switch the docker install instruction to one proposed in https://docs.docker.com/engine/install/ubuntu/

integration_tests/scripts/run_local_docker_tests.sh
```diff
@@ -0,0 +1,471 @@
...
+      ;;
+  esac
+done
+
+
+# ===== VALIDATION =====
+validate_paths() {
+  step "Validating paths"
+
+  # FIX 1: Correct secret folder path (was wrong in original — data/sample_v3_values doesn't exist)
```
ajasnosz: remove the comment

integration_tests/scripts/run_local_docker_tests.sh
```diff
@@ -0,0 +1,471 @@
...
+# ===== CREATE INDEXES =====
+create_indexes() {
+  step "Creating Splunk indexes"
+  # FIX 2: Use exact index names the tests search against
+  declare -A INDEXES=(
+    [netmetrics]="metric"
+    [em_metrics]="metric"
+    [netops]="event"
+    [em_events]="event"
+    [em_meta]="event"
```
ajasnosz: I think we might not need this index, as it can be some old dependencies. Could you check if tests really are using this index?

integration_tests/scripts/run_local_microk8s_tests.sh
```diff
@@ -0,0 +1,517 @@
...
+    | sudo tee /etc/apt/sources.list.d/docker.list > /dev/null
+
+  sudo apt-get update
+
+  # Check available versions
+  apt-cache madison docker-ce | grep "29.2.1"
+
+  # Install exact version
+  sudo apt-get install -y \
+    docker-ce=5:29.2.1-1~ubuntu.22.04~jammy \
```
ajasnosz: why do we need the specific docker version?

integration_tests/scripts/run_local_microk8s_tests.sh
```diff
@@ -0,0 +1,517 @@
...
+  else
+    info "Starting Splunk in Docker..."
+    mkdir -p "$PWD/splunk-data"
+    sudo docker run -d \
+      -p 8000:8000 \
+      -p 8088:8088 \
+      -p 8089:8089 \
+      -e SPLUNK_GENERAL_TERMS="--accept-sgt-current-at-splunk-com" \
+      -e SPLUNK_START_ARGS="--accept-license" \
+      -e SPLUNK_PASSWORD="changeme2" \
```
ajasnosz: in line 56 we are setting the default param for password, but here we are still assigning it in string and not from variable.

integration_tests/tests/test_poller_integration.py
```diff
@@ -1934,11 +1932,22 @@ def test_wrong_groups(self, request, setup_splunk):
...
-def run_retried_single_search(setup_splunk, search_string, retries):
-    for _ in range(retries):
+# def run_retried_single_search(setup_splunk, search_string, retries):
+#     for _ in range(retries):
+#         result_count, metric_count = splunk_single_search(setup_splunk, search_string)
+#         if result_count or metric_count:
+#             return result_count, metric_count
+#         logger.info("No results returned from search. Retrying in 2 seconds...")
+#         time.sleep(2)
+#     return 0, 0
```
ajasnosz: remove commented code

Makefile
```diff
@@ -1,6 +1,103 @@
...
+k8s-all:
+	bash $(MICROK8S_SCRIPT) --clean --test all
+
+
+# ============================================
+# Utilities
+# ============================================
+
+.PHONY: setup
+setup:
```
ajasnosz: I would change the name to something that would be more explicit to the task done here.

ajasnosz (changes requested):

integration_tests/scripts/run_local_microk8s_tests.sh
pkalevar16: We are installing a specific Docker version (docker-ce=5:29.2.1-1~ubuntu.22.04~jammy) because the script internally runs the command:

os.system("sudo docker compose up -d")

This command relies on the Docker Compose V2 plugin which is bundled and works correctly with this Docker version.

If a different or older Docker version is installed, the docker compose command may not be available or may behave differently (for example requiring docker-compose instead).

To avoid compatibility issues during integration test execution, we pinned the Docker version.

If we remove the version pin, we may need to modify the script to handle both cases (e.g., detect and switch between docker compose and docker-compose

integration_tests/scripts/run_local_docker_tests.sh
pkalevar16: yes

integration_tests/scripts/automatic_setup_compose.sh
ajasnosz: is this commented line needed?

integration_tests/scripts/run_local_docker_tests.sh
ajasnosz: Where do we use this index? Do we see some data coming to this index when running tests?

integration_tests/scripts/run_local_docker_tests.sh
```diff
@@ -0,0 +1,513 @@
...
+    error "Cannot connect to Splunk at ${SPLUNK_API}"
+    error "Check --host and --password args"
+    exit 1
+  fi
+}
+
+# ===== CREATE INDEXES =====
+create_indexes() {
+  step "Creating Splunk indexes"
+  # FIX 2: Use exact index names the tests search against
```
ajasnosz: remove comment

integration_tests/scripts/run_local_microk8s_tests.sh
ajasnosz: I think if we are using the same method as in docker link to install on ubuntu the version should always be the latest similar to installing the docker compose v2. 
Can we confirm which version would be installed then?

ajasnosz (changes requested):

Makefile
```diff
@@ -1,6 +1,103 @@
...
 	rm -rf rendered/manifests/tests/splunk-connect-for-snmp/charts
 	./render_manifests.sh
+
+
+# ============================================
+# Docker Integration Tests
+# ============================================
+
+.PHONY: docker-traps
+docker-traps:
```
omrozowicz-splunk: I think we should change all the names to be prefixed with `test-*` as this is confusing for someone who doesn't know the code to have `docker-traps` in a main Makefile. I'd say this is something that spin up the docker traps service.

omrozowicz-splunk (approved):

ajasnosz (approved):

PR #1356, opened 2026-03-13
feat: add max oid to process

ajasnosz:
# Description

Adds limits for getBulk and getCmd calls to split the amount of oids asked per one request. The max_oid_to_process flag would be also set per device in inventory. 

Fixes # (issue)

## Type of change

- [ ] New feature

## How Has This Been Tested?

manual, unit

## Checklist


splunk_connect_for_snmp/snmp/manager.py
```diff
@@ -422,39 +429,46 @@ def run_bulk_request(
...
-            MAX_REPETITIONS,
-            *varbinds_bulk,
-            lexicographicMode=False,
-            ignoreNonIncreasingOid=is_increasing_oids_ignored(ir.address, ir.port),
+        for varbind_chunk in self.get_varbind_chunk(
+            list(varbinds_bulk), max_oid_to_process
         ):
-            if not _any_failure_happened(
+            logger.info(
+                f"Running bulk request for {address} with max_oid_to_process {max_oid_to_process} with varbinds {varbind_chunk}"
```
omrozowicz-splunk: shouldn't it be on debug level? 🤔

omrozowicz-splunk (commented):
so we're putting it in our csv structure now? Wouldn't be better to keep it separate in additional part of the configuration, like:

```
poller:
  maxOid:
    - ip: 1.1.1.1
    - limit: 20
```

bacuase if we modify `inventory.csv` that users already have, it can be considered a breaking change

ajasnosz:
@omrozowicz-splunk I think keeping it in this structure would be more consistent with how we define changes per device. It is backward compatible so if user don't add this column to inventory the code would still work with assigning the max_oid from global variable.

splunk_connect_for_snmp/snmp/manager.py
ajasnosz: yes, or I can remove it? I had this log added for local testing

omrozowicz-splunk: remove also seems okay

omrozowicz-splunk:
> @omrozowicz-splunk I think keeping it in this structure would be more consistent with how we define changes per device. It is backward compatible so if user don't add this column to inventory the code would still work with assigning the max_oid from global variable.

it works as well with groups? when we have different settings per device in a group?

ajasnosz:
It will work with groups in both ways. Either from setitng in inventory, or specify per device in the group:
```
  groups: |
    example_group_1:
      - address: 172.31.19.169
        port: 1166
        max_oid_to_process: 3
```

omrozowicz-splunk (approved):

PR #1373, opened 2026-03-30
chore: import issue on ui tests

ajasnosz:
# Description

Fixing error from ui test:
`ModuleNotFoundError: No module named 'splunk_cim_models'`

Removes pytest-splunk-addon library that was used only for adding the splunk flags to CLI options from 
conftest.py

Fixes # (issue)

## Type of change

- [x] Bug fix

## How Has This Been Tested?

ui tests

## Checklist

### DIFF
--- a/ui_tests/conftest.py
+++ b/ui_tests/conftest.py
@@ -17,6 +17,27 @@ def pytest_addoption(parser):
         default="127.0.0.1",
         help="Device Simulator external IP",
     )
+    parser.addoption(
+        "--splunk-host",
+        action="store",
+        dest="splunk-host",
+        default="localhost",
+        help="Splunk host to connect to",
+    )
+    parser.addoption(
+        "--splunk-user",
+        action="store",
+        dest="splunk-user",
+        default="admin",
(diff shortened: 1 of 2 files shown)


ui_tests/conftest.py
```diff
@@ -17,6 +17,27 @@ def pytest_addoption(parser):
         default="127.0.0.1",
         help="Device Simulator external IP",
     )
+    parser.addoption(
+        "--splunk-host",
+        action="store",
+        dest="splunk-host",
+        default="localhost",
```
Kawron: I would use 127.0.0.1, sometimes localhost may not be working e.g. if someone fiddled with /etc/hosts

Kawron: And we use ip in --device-simulator

Kawron (changes requested):

ui_tests/conftest.py
ajasnosz: done

Kawron (approved):

PR #1422, opened 2026-05-28
feat(traps): improve varbind MIB resolution and add decode limits

stbakercs:
# Description
## Summary

This PR improves SNMP trap varbind handling across the trap receiver and trap worker: more reliable MIB resolution (including NOTIFICATION-TYPE and table-instance OIDs), safer ASN.1 serialization for the Celery queue, an optional path to retain untranslated varbinds in Splunk events, and a configurable cap on how many varbinds are decoded per trap to protect CPU and memory under large traps.

---

## Problem

- Trap varbinds with plain Python scalars, NOTIFICATION-TYPE OIDs (e.g. `snmpTrapOID`), or table-instance numeric OIDs often failed MIB resolution or produced incomplete Splunk events.
- OCTET STRING values (e.g. 4-byte IPv4 addresses) sometimes serialized poorly via `prettyPrint()`, blocking downstream casting.
- Very large traps could drive heavy MIB lookup, re-resolution, and queue payload growth with no upper bound.
- Varbinds that resolved in the internal table but were skipped by `process_snmp_data` could be dropped entirely.

---

Fixes # (issue)
### Trap worker (`splunk_connect_for_snmp/snmp/tasks.py`)

- Refactored `trap()` into explicit phases: limit → preload MIBs → resolve → re-resolve placeholders → process metrics → optional unresolved append.
- **`_resolve_trap_varbind`**: `ObjectIdentity` resolution with `getNodeLocation` fallback for table instances; NOTIFICATION-TYPE handling via `_object_type_from_resolved_identity`.
- **`_coerce_trap_varbind_value`**: Maps plain ints/strings/bools to ASN.1 types expected by pysnmp using MIB syntax when available.
- **`_preload_trap_mibs`**: Loads enterprise MIBs (`1.3.6.1.4.1.*`) before first resolution when referenced in varbind names/values.
- **`_process_trap_metrics`**: Bounded retry loop (`_TRAP_MIB_RETRY_MAX = 5`) when `process_snmp_data` requests more MIBs; full re-resolve after new MIB loads.
- **`_re_resolve_trap_work_if_needed`**: Second full pass when varbinds still show placeholder symbols (e.g. `enterprises.*`).
- **`INCLUDE_UNRESOLVED_TRAP_VARBINDS`** (default `false`): Adds `sc4snmp::unresolved` fields for failed or unprocessed varbinds.
- **`MAX_TRAP_VARBINDS_TO_DECODE`**: Applied again on the worker after queue delivery (defense in depth).

[...]

PR #1423, opened 2026-06-02
feat(traps): improve varbind MIB resolution and add decode limits 

ajasnosz:
More description in pr #1422 

Improve trap varbind handling end-to-end: stronger MIB resolution on the worker, safer ASN.1 serialization at the trap receiver, optional forwarding of untranslated varbinds, and a configurable cap on varbinds decoded per trap. Trap worker (tasks.py):
- Refactor trap() with dedicated resolve, preload, retry, and re-resolve paths
- Coerce plain Python values to ASN.1 (NOTIFICATION-TYPE, integers, MIB syntax)
- Resolve table-instance OIDs via getNodeLocation decomposition
- Preload enterprise MIBs (1.3.6.1.4.1.*) before translation
- Bounded MIB load/re-resolve loop; track loaded MIBs via load_mibs() return value
- Optional INCLUDE_UNRESOLVED_TRAP_VARBINDS → sc4snmp::unresolved fields
- Apply MAX_TRAP_VARBINDS_TO_DECODE before processing Trap receiver (traps.py, manager.py):
- format_trap_varbind_value(): dotted IPv4 for 4-octet OCTET STRINGs, hex fallback
- Limit varbind pairs at receive time (shared trap_varbind_limit module) Configuration and docs:
- MAX_TRAP_VARBINDS_TO_DECODE (1–500, default 25n traps receiver and worker
- INCLUDE_UNRESOLVED_TRAP_VARBINDS for worker-trap
- Helm values, schema, templates; Docker Compose .env and compose services
- Document in traps.md, workers.md, docker compose env reference, microk8s values Tests:
- New test_trap_varbind_limit.py
- Expanded trap resolution/coercion/retry coverage in test_tasks.py
- Updates to test_do_work, test_get_varbinds, test_mibs

# Description

Fixes # (issue)

## Type of change

- [ ] Dependency update
- [ ] Bug fix
- [ ] New feature
- [ ] Refactor/improvement

## How Has This Been Tested?

## Checklist


splunk_connect_for_snmp/snmp/manager.py
```diff
@@ -194,6 +194,28 @@ def get_group_key(mib, oid, index) -> str:
...
+
+    ``prettyPrint()`` returns an empty string or hex for some OCTET STRING payloads
+    (including InetAddressIP). Convert 4-octet values to dotted IPv4 so MIB
+    resolution can cast them.
+    """
+    with suppress(AttributeError, TypeError):
+        octets = bytes(val.asOctets())
+        if len(octets) == 4:
+            return ".".join(str(b) for b in octets)
+    displayed = val.prettyPrint()
```
omrozowicz-splunk: When exactly prettyprint returns None? I thought it is either value or raising an error

splunk_connect_for_snmp/snmp/manager.py
```diff
@@ -194,6 +194,28 @@ def get_group_key(mib, oid, index) -> str:
 MTYPES = tuple(["cc", "c", "g"])
 
 
+def format_trap_varbind_value(val) -> str:
+    """
+    Serialize a trap varbind ASN.1 value for the Celery worker queue.
+
+    ``prettyPrint()`` returns an empty string or hex for some OCTET STRING payloads
+    (including InetAddressIP). Convert 4-octet values to dotted IPv4 so MIB
```
omrozowicz-splunk: why only ipv4 and no ipv6?

charts/splunk-connect-for-snmp/values.schema.json
```diff
@@ -527,6 +527,20 @@
...
             },
+            "includeUnresolvedVarbinds": {
+              "type": "object",
+              "additionalProperties": false,
+              "properties": {
+                "enabled": {
+                  "type": "boolean"
+                }
+              }
+            },
```
ajasnosz: I think this is an unneeded structure, as there won't be any other object under this value, we can just do it as boolean and eventually rename to `enableIncludeUnresolvedVarbinds`

charts/splunk-connect-for-snmp/values.schema.json
```diff
@@ -527,6 +527,20 @@
...
+            "maxVarbindsToDecode": {
+              "type": "integer",
+              "minimum": 1,
+              "maximum": 500
+            },
```
ajasnosz: move maximum and defauult value to be unlimited for backward compatibility

charts/splunk-connect-for-snmp/values.schema.json
```diff
@@ -826,6 +840,11 @@
         "discoverEngineId": {
           "type": ["string", "boolean"]
         },
+        "maxVarbindsToDecode": {
+          "type": "integer",
+          "minimum": 1,
+          "maximum": 500
+        },
```
ajasnosz, line 835: The values for traps even if needed in both worker and trap, should be the same during runtime, so I would rather define them in one place for user to specify, and then copy from that one value to both pods/containers

splunk_connect_for_snmp/snmp/tasks.py
```diff
@@ -146,81 +181,400 @@ def resolve_address(address: str):
...
+    if isinstance(value, (int, float)):
+        return snmp_rfc1902.Integer(int(value))
+    if isinstance(value, str):
+        if value == "":
+            return value
+        syntax_name = getattr(syntax, "__name__", "") if syntax is not None else ""
+        if syntax_name in _STRING_SYNTAX_NAMES:
+            return value
+        if syntax_name in _INTEGER_SYNTAX_NAMES and _INTEGER_STRING.match(value):
+            return snmp_rfc1902.Integer(int(value))
```
omrozowicz-splunk: what if the integer value goes out of scope of Integer (I think it is range-constrained [-2147483648, 2147483647] so we will never get a Counter64. But this is to be checked

splunk_connect_for_snmp/snmp/tasks.py
```diff
@@ -146,81 +181,400 @@ def resolve_address(address: str):
...
-@shared_task(bind=True, base=Poller)
-def trap(self, work):
-    varbind_table, not_translated_oids, remaining_oids, remotemibs = [], [], [], set()
-    metrics = {}
-    work["host"] = format_ipv4_address(work["host"])
+def _oids_for_mib_lookup(varbind):
+    """Return numeric OID strings from a trap varbind name and/or value."""
+    name, value = varbind
+    for candidate in (name, value):
+        if OID_VALIDATOR.match(candidate):
```
omrozowicz-splunk, line 175: can candidate be something else than string? if it's something else than string this method will crash

splunk_connect_for_snmp/snmp/tasks.py
```diff
@@ -146,81 +181,400 @@ def resolve_address(address: str):
...
+    result = {}
+    for _ in range(_TRAP_MIB_RETRY_MAX):
+        retry, remotemibs, result = self.process_snmp_data(varbind_table, metrics, host)
+        if not retry or not remotemibs:
+            break
+        if _load_new_trap_mibs(self, remotemibs):
+            varbind_table, _ = _resolve_work_varbinds(self, work_data)
+        else:
+            # MIB names are known and already loaded; re-resolve once and re-process.
+            varbind_table, _ = _resolve_work_varbinds(self, work_data)
```
omrozowicz-splunk: does the else condition make sense? both if and else calls _resolve_work_varbinds, at the beginning of the loop process_snmp_data is done so the behaviour is very similar. Also if something new fails in `_, _, result = self.process_snmp_data(varbind_table, metrics, host)` it will be silently lost

splunk_connect_for_snmp/snmp/tasks.py
```diff
@@ -146,81 +181,400 @@ def resolve_address(address: str):
...
         try:
-            varbind_table.append(
-                ObjectType(ObjectIdentity(w[0]), w[1]).resolveWithMib(
-                    self.mib_view_controller
-                )
-            )
+            return syntax.clone(value)
+        except Exception as exc:
+            logger.debug("MIB syntax clone failed for %s: %s", mib_node, exc)
+    if isinstance(value, bool):
```
omrozowicz-splunk: not sure if this is needed it works the same way as int (redundant?)

splunk_connect_for_snmp/snmp/tasks.py
```diff
@@ -146,81 +181,400 @@ def resolve_address(address: str):
...
         except SmiError:
-            logger.warning(f"No translation found for {w[0]}")
+            logger.warning(f"No translation found for {name}")
+            unresolved.append((name, value))
+    return varbind_table, unresolved
+
+
+@shared_task(bind=True, base=Poller)
+def trap(self, work):
+    log_trap_varbind_limit_config(logger)
```
omrozowicz-splunk: shouldn't it be logged only on startup?

omrozowicz-splunk (changes requested):

splunk_connect_for_snmp/snmp/tasks.py
```diff
@@ -49,7 +61,30 @@
...
+    {"OctetString", "Opaque", "IpAddress", "DisplayString", "PhysAddress"}
+)
+_INTEGER_SYNTAX_NAMES = frozenset(
+    {
+        "Integer",
+        "Integer32",
+        "Counter32",
+        "Counter64",
+        "Gauge32",
+        "Unsigned32",
```
ajasnosz: missing Unsigned64 and Gauge64

stbakercs:
I have addressed the comments, working on the process to get them committed back to this feature branch.

stbakercs:
Since the branch changed git just hangs and errors when I try to work with it I had to recreated the fork and this branch.  You above comments are addressed in the new fork https://github.com/splunk/splunk-connect-for-snmp/pull/1424

splunk_connect_for_snmp/snmp/manager.py
stbakercs: prettyPrint: Clarified in docstring; guard handles empty str and rare non-str; no behavior change unless we drop the guard.

splunk_connect_for_snmp/snmp/manager.py
stbakercs: Support added  IPv6: 16-octet OCTET STRINGs formatted via inet_ntop

splunk_connect_for_snmp/snmp/tasks.py
stbakercs: Added Integer/Counter64: Coercion follows MIB syntax name; large values use Counter64

splunk_connect_for_snmp/snmp/tasks.py
stbakercs: updated instance check

splunk_connect_for_snmp/snmp/tasks.py
stbakercs: Simplified the checks

splunk_connect_for_snmp/snmp/tasks.py
stbakercs: rewoked the logic

splunk_connect_for_snmp/snmp/tasks.py
stbakercs: Correct the logic

charts/splunk-connect-for-snmp/values.schema.json
stbakercs: Removed

charts/splunk-connect-for-snmp/values.schema.json
stbakercs: reworked this area of the config

charts/splunk-connect-for-snmp/values.schema.json
stbakercs: Updated

splunk_connect_for_snmp/snmp/tasks.py
stbakercs: Address the missing items

stbakercs:
This PR can be closed due to https://github.com/splunk/splunk-connect-for-snmp/pull/1424

splunk_connect_for_snmp/snmp/tasks.py
omrozowicz-splunk: I'm not 100% sure about this comment as I haven't worked with traps for a while, but when candidate is not a string shoudn't we cast it to be a string or log an error? I suspect it should be a string most of the time (unless something unexpected happens)

splunk_connect_for_snmp/snmp/tasks.py
omrozowicz-splunk: thanks! One more thing - what happens when the value is within the int range but is a float? Like 3.9? Will it be rounded?

splunk_connect_for_snmp/snmp/tasks.py
omrozowicz-splunk: do I understand it right that we merge those previous `unresolved` data with the new `still_unresolved` result? Is this what we want? the merge function does "first one wins" strategy and I think we should prefer `still_unresolved` is it is the most fresh and should ultimately replace previous `unresolved` imo

omrozowicz-splunk: also dedupes by name sounds problematic as well, what if the same OID can appear multiple times? First value stays, the other ones are dropped

ajasnosz:
Changes added in my last 2 commits:

**Fix some incorrectly unresolved values**
When sending traps, varbinds with OIDs that should be resolvable (e.g. 1.3.6.1.2.1.2.2.1.8.1 = ifOperStatus.1) were showing in Splunk as unresolved::1.3.6.1.2.1.2.2.1.8.1 instead of being translated.

Trap values arrive as plain strings (from prettyPrint()), and the code decided how to coerce them by matching the syntax's class name against hardcoded sets like _INTEGER_SYNTAX_NAMES. But MIBs define many fields as named TEXTUAL-CONVENTIONs, so pysnmp compiles them into subclasses with names like IfOperStatus, TruthValue, or vendor HC counters — none of which are in those sets. The name match failed, value coercion fell through to return value (the raw string), and resolveWithMib() then rejected it → the OID landed in unresolved. This affected any named integer/counter/gauge subtype.

**Solution**
Switched from name-based matching to inheritance-based classification:
- Added _syntax_base_category() — uses issubclass against pyasn1's univ.Integer / univ.OctetString, the common ancestors of all SNMP numeric and string types (including every named TC subtype).
- Rewrote _coerce_trap_varbind_value() — converts the string to the syntax's base Python type, then clones from the actual MIB syntax object, preserving the exact subtype (named values, 64-bit width, etc.). Falls back gracefully for constraint violations.

**Fix IpAddress format**

- format_trap_varbind_value now formats IP by ASN.1 type, not octet length (genuine IpAddress → dotted; bare octet strings → IPv4/IPv6 only when bytes look like a raw address via new _looks_like_raw_address, else hex/text), using ipaddress/rfc1902.
- _asn1_from_numeric is range-safe (Counter64 only for 0..2**64-1, otherwise returns text instead of crashing); _merge_unresolved_trap_varbinds dedups by exact (name, value) pair and the re-resolve/retry paths return the fresh unresolved set; _load_new_trap_mibs tracks attempted MIBs so a failing MIB isn't retried every trap.

omrozowicz-splunk (approved):

PR #1443, opened 2026-07-06
feat: support of python3.13, lextudios pysnmp and multi-bulk-walk-cmd…

gmaiyani:
# Description

Adding support of python v3.13 and migrating from pysnmplib to lextudio's pysnmp v7.1.27

- Replaced pysnmplib with lextudio’s pysmp v7.1.27 (current latest version)
- Replaced pysnmp-pysmi to lextudio's pysmi
- Included pysnmpcrypto (maintained by lextudio's) for stronger crypto algorithms with pysnmp v6+
- Made SC4SNMP compatible with Python 3.13
- Updated SNMP walk implementation from bulkCmd to custom multi_bulk_walk_cmd
- Introduced deploymentUser in values.yaml to make the sc4snmp entirely compatible with lextudio's pysnmp.
- Added DEPLOYMENT_USER variable for Docker Compose deployment.
- Introduces multi_bulk_walk_cmd to walk multiple OID trees simultaneously with lexicographic mode control per varbind.
- Migrate auto-discovery feature to make compitable with lextuio's pysnmp and python 3.13
- Replace the thread-pool executor with semaphore asyncio for concurrency control

- [x] Dependency update
- [x] New feature
- [x] Refactor/improvement
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist


docs/microk8s/configuration/deployment-configuration.md
```diff
@@ -11,6 +11,20 @@ microk8s helm3 inspect values splunk-connect-for-snmp/splunk-connect-for-snmp >
 
 The whole file is divided into the following parts:
 
+To configure `deploymentUser`:
+
+- The `deploymentUser` configuration is kept in the `values.yaml` file as a global (top-level) parameter.
+- `values.yaml` is used during the installation process for configuring Kubernetes values.
```
omrozowicz-splunk, line 17: Is it also necessary for docker installation?

gmaiyani: > Is it also necessary for docker installation?

Yes it is necessary in both deployment

omrozowicz-splunk (approved):

docs/microk8s/configuration/deployment-configuration.md
```diff
@@ -11,6 +11,20 @@ microk8s helm3 inspect values splunk-connect-for-snmp/splunk-connect-for-snmp >
 
 The whole file is divided into the following parts:
 
+To configure `deploymentUser`:
+
+- The `deploymentUser` configuration is kept in the `values.yaml` file as a global (top-level) parameter.
```
ajasnosz, line 16: it should be also mentioned in table in values-params-description.md

docs/microk8s/configuration/deployment-configuration.md
ajasnosz: If it's necessary in both and has similar explanation would it be possible to move the description somewhere under docs/configuration instead of microk8s?

splunk_connect_for_snmp/traps.py
```diff
@@ -428,13 +433,13 @@ def main():
...
-                config.addV3User(
+                config.add_v3_user(
                     snmp_engine,
                     userName=username,
                     authProtocol=auth_protocol,
-                    authKey=auth_key,
+                    authKey=auth_key if auth_key else None,
                     privProtocol=priv_protocol,
-                    privKey=priv_key,
+                    privKey=priv_key if priv_key else None,
```
ajasnosz, line 442: do we need here the None, isn't it by defaul assigned when defining the variable with get_secret_value?

splunk_connect_for_snmp/snmp/tasks.py
```diff
@@ -593,7 +604,7 @@ def _retry_trap_varbinds_after_mib_load(
 
 @shared_task(bind=True, base=Poller)
 def trap(self, work):
-    metrics = {}
+    metrics = {}  # type: ignore
```
ajasnosz, line 607: is it some comment for linter?

splunk_connect_for_snmp/snmp/auth.py
```diff
@@ -77,14 +77,12 @@ def get_security_engine_id(logger, rt: RecordType, snmp_engine: SnmpEngine):
...
 
-    error_indication, _, _, _ = next(
-        getCmd(
-            snmp_engine,
-            auth_data,
-            transport_target,
-            ContextData(),
-            ObjectType(ObjectIdentity("SNMPv2-MIB", "sysDescr", 0)),
-        )
+    error_indication, _, _, _ = await get_cmd(
```
ajasnosz: did the data type changed so we don't need next() anymore?

splunk_connect_for_snmp/snmp/manager.py
```diff
@@ -370,35 +373,52 @@ def get_snmp_engine(self, version="", create_new=False) -> SnmpEngine:
         """
         if version == "3" or create_new:
             snmp_engine = SnmpEngine()
-            snmp_engine.setUserContext(mibViewController=self.mib_view_controller)
+            # snmp_engine.set_user_context(mibViewController=self.mib_view_controller)
```
ajasnosz: we can remove the comment, as we move with new cache approach

splunk_connect_for_snmp/snmp/multi_bulk_walk_cmd.py
```diff
@@ -0,0 +1,328 @@
...
+      OID subtrees in parallel within a single SNMP session.
+     - Each varbind is independently tracked: completion, last OID, and
+      lexicographic boundaries are maintained per varbind.
+     - lexicographicMode is applied per-varbind: a varbind stops only when its
+      own scope is exhausted if lexicographicMode=False. Other varbinds continue.
+     - Can optionally ignore non-increasing OIDs: if ignoreNonIncreasingOid=True,
+      the generator continues even when the agent returns OIDs that are not
+      greater than previous ones.
+     - Returns all valid responses from each GETBULK call in a single yield: all
+      active varbinds' results are grouped together in the same batch.
```
ajasnosz: is ' a typo?

splunk_connect_for_snmp/discovery/discovery_manager.py
```diff
@@ -29,108 +38,207 @@ def get_host_list(self, subnet):
...
-                    else:
-                        logger.info(f"SNMP not enabled on the device: {ip}")
-                except Exception as e:
-                    logger.error(
-                        f"Snmp check for device {ip} generated an exception : {e}"
+        if error_status != 0:
+            logger.debug(
+                f"SNMP error status for {ip}: {error_status} at index {error_index}"
+            )
+            return None
```
ajasnosz, line 129: why do we separate those scenarios? Doesn't error_indication means that there will be error_status?

splunk_connect_for_snmp/discovery/discovery_manager.py
```diff
@@ -1,24 +1,33 @@
...
 from splunk_connect_for_snmp.common.discovery_record import DiscoveryRecord
+from splunk_connect_for_snmp.discovery.exceptions import DiscoveryError
 from splunk_connect_for_snmp.snmp.auth import get_auth, setup_transport_target
 
 logger = get_task_logger(__name__)
 
 DISCOVERY_FOLDER_PATH = os.getenv("DISCOVERY_FOLDER_PATH", "/app/discovery")
 DISCOVERY_CSV_PATH = os.path.join(DISCOVERY_FOLDER_PATH, "discovery_devices.csv")
 DISCOVERY_LOCK_PATH = os.path.join(DISCOVERY_FOLDER_PATH, "discovery_devices.lock")
+DEFAULT_CONCURRENCY = 10
```
ajasnosz: do we want to hardcode this value and not set it from the user config?

splunk_connect_for_snmp/discovery/discovery_manager.py
```diff
@@ -29,108 +38,207 @@ def get_host_list(self, subnet):
...
+                logger.debug(
+                    f"SNMP device found for {ip_list[idx]}: {result}. Device is from discovery: {discovery_record.discovery_name}"
+                )
+                devices_detail.append(result)
 
         return devices_detail
 
     def add_devices_detail_to_csv(
-        self, snmp_devices_detail, delete_flag, discovery_name
+        self, snmp_devices_detail, delete_flag, dicovery_name
```
ajasnosz: typo in the variable name should be discovery_name for the whole file

ajasnosz (changes requested):

splunk_connect_for_snmp/snmp/tasks.py
gmaiyani: yes it is for linter

splunk_connect_for_snmp/snmp/auth.py
gmaiyani: yes, return type is changed so we do not need next

splunk_connect_for_snmp/snmp/manager.py
gmaiyani: Yes, i forgot to remove

splunk_connect_for_snmp/snmp/multi_bulk_walk_cmd.py
gmaiyani: yes it is, i will correct it

splunk_connect_for_snmp/discovery/discovery_manager.py
gmaiyani: I think error_indication indicates a request timeout  or unreachable target or an authentication or transport problem, or another error that prevents pysnmp from completing the request and receiving a valid response. on the other hand, error_status indicates that the target device responded, but its snmp agent could not complete the requested operation. so for the exact logging purpose we have seperated those things

splunk_connect_for_snmp/discovery/discovery_manager.py
gmaiyani: We can provide it from the user config. let me know if we need to go with that approach

splunk_connect_for_snmp/traps.py
gmaiyani: Not really as per the logic but since the return type of get_secret_value is string it would be good to keep this.
what do you suggest?

docs/microk8s/configuration/deployment-configuration.md
gmaiyani: yea, but since doc -> configuration are more specific to the feature usage i don't know where we can put this And for docker we already have separate the environment variable section so it would be best fit for the docker but not sure about the microk8s.

docker_compose/.env
```diff
@@ -10,6 +10,7 @@ LOCAL_MIBS_PATH="./local_mibs"
 SECRET_FOLDER_PATH="./secrets"
 DISCOVERY_PATH="./discovery"
 SC4SNMP_VERSION="1.17.1-beta.1"
+DEPLOYMENT_USER=
```
ajasnosz: maybe it would be better to also assign here the default?

splunk_connect_for_snmp/discovery/discovery_manager.py
ajasnosz: I think that would be a better option, as customer can have different env and network possibilities.

splunk_connect_for_snmp/traps.py
ajasnosz: Ok, we can keep it

docs/microk8s/configuration/deployment-configuration.md
ajasnosz: You're right we don't have the common page like that, so it can stay per section, but then please add similar explanation to the value in the docker installation, as I think it was only mentioned in the table.

lextm:
You should use pyasn1/pyasn1. Please don't use LeXtudio/pyasn1, as it is out of date.

gmaiyani:
> You should use pyasn1/pyasn1. Please don't use LeXtudio/pyasn1, as it is out of date.

Sure, thanks for the update

ajasnosz (approved):

ajasnosz (changes requested):
I see after merge some of the links in documentation got swapped to the wrong ones.

ajasnosz (approved):

PR #1451, opened 2026-07-14
test: integration test cases for auto-discovery

gmaiyani:
# Description

- Deploys autodiscovery SNMP simulators through MicroK8s for both SC4SNMP deployment types.
Uses separate namespaces and IP ranges:MicroK8s: 10.1.1.0/29 and 10.2.2.0/29
Docker Compose: 10.4.4.0/29 and 10.5.5.0/29

- Limits simulator and discovery-service setup to the part7 matrix job.
Fixes # (issue)

## Type of change

- [ ] Dependency update
- [ ] Bug fix
- [ ] New feature
- [ ] Refactor/improvement

## How Has This Been Tested?

## Checklist


integration_tests/scripts/automatic_setup_microk8s.sh
```diff
@@ -1,17 +1,21 @@
-#!/bin/bash
+#!/usr/bin/env bash
+set -euo pipefail
```
ajasnosz: I think we can remove the pipefail

integration_tests/scripts/automatic_setup_microk8s.sh
```diff
@@ -30,6 +34,68 @@ function yellow {
...
+      consecutive_successes=0
+      echo $(green "[INFO] MicroK8s API is not ready (attempt ${attempt}); retrying")
+    fi
+    sleep 3
+  done
+
+  echo $(red "MicroK8s API did not remain ready for five consecutive checks")
+  return 1
+}
+
```
ajasnosz: I think we already had similar thing implemented, to return the logs and microk8s get pods or docker containers when the tests would fail. Do we need to reimplement it for the autodiscovery or are those logs incomplete? Then I think we can add it for all tests.

ajasnosz (changes requested):

integration_tests/scripts/automatic_setup_microk8s.sh
gmaiyani: Yes, but that will be used only when there are any failure during the test execution, on other hand this will run when we have any failure during environment installation and deployment (such as deploying agent simulator or docker/microk8s deployment). may be we can keep this in separate file (diagnostic.sh )or we can put this into _common.sh) or we can keep it as its ?

ajasnosz (approved):

PR #1466, opened 2026-07-24
fix: prevent duplicate profiles when merging varbind mappings

gmaiyani:
# Description

- Fixes unbounded growth of profile labels in `get_mapping` and `bulk_mapping` when multiple profiles containing the same OID are combined repeatedly in different orders.

- The existing merge logic modifies cached profile mappings and compares complete comma-separated strings instead of individual profile names. This can cause duplicate profile labels to grow at exponential rate inside a long-running Celery poller child.

Fixes # (issue)

- Prevent duplicate profiles when merging varbind mappings to avoid mappings growing unnecessarily large

## Type of change

- [x] Bug fix

## How Has This Been Tested?

## Checklist

### DIFF
--- a/splunk_connect_for_snmp/snmp/varbinds_resolver.py
+++ b/splunk_connect_for_snmp/snmp/varbinds_resolver.py
@@ -221,19 +221,61 @@ class Profile:
 
     def add_mappings(self, dict1, dict2) -> dict:
         """
-        Helper to adding varbinds mapping dictionaries inside profiles.
-        When there are the same keys in both dictionaries they will be added after comma.
-        :param dict1:
-        :param dict2:
-        :return mapping:
+        Merge two OID-to-profile mappings without modifying either input.
+
+        Each dictionary maps an OID to one or more comma-separated profile
+        names. When an OID exists in both dictionaries, each distinct profile
+        name is retained once, in first-seen order.
+
+        The previous logic compared the full comma-separated profile string,
+        so duplicate profiles could be added when the profile order differed.
+
+        :param dict1: Base OID-to-profile mapping.
+        :param dict2: OID-to-profile mapping to merge into the base mapping.
+        :return: A new merged mapping containing distinct profile names.
         """
-        mapping = dict1
-        for k, v in dict2.items():
-            if mapping.get(k) and v not in mapping.get(k):
-                mapping[k] = f"{mapping[k]},{v}"
-            elif not mapping.get(k):
-                mapping[k] = v
+        mapping = {
+            mapping_key: ",".join(self.unique_profile_names(profile_names))
+            for mapping_key, profile_names in dict1.items()
+        }
+
+        for mapping_key, profile_names in dict2.items():
+            dict1_profiles = self.unique_profile_names(mapping.get(mapping_key, ""))
+            dict2_profiles = self.unique_profile_names(profile_names)
+
+            dict1_profile_set = set(dict1_profiles)
+            added_profiles = [
+                profile_name
+                for profile_name in dict2_profiles
+                if profile_name not in dict1_profile_set
+            ]
+            result_profiles = dict1_profiles + added_profiles
+            mapping[mapping_key] = ",".join(result_profiles)
+
+            logger.debug(
+                f"Profile.add_mappings name={self.name} key={mapping_key} "
+                f"dict1_profiles={dict1_profiles} "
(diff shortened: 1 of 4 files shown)


splunk_connect_for_snmp/snmp/varbinds_resolver.py
```diff
@@ -221,20 +221,62 @@ def __repr__(self):
...
-            elif not mapping.get(k):
-                mapping[k] = v
+        mapping = {
+            mapping_key: ",".join(self.unique_profile_names(profile_names))
+            for mapping_key, profile_names in dict1.items()
+        }
+
+        for mapping_key, profile_names in dict2.items():
+            dict1_profiles = self.unique_profile_names(mapping.get(mapping_key, ""))
+            dict2_profiles = self.unique_profile_names(profile_names)
```
ajasnosz: can we name the dic1 and dict2 something more meaningful like base / incoming mapping?

splunk_connect_for_snmp/snmp/varbinds_resolver.py
```diff
@@ -221,20 +221,62 @@ def __repr__(self):
...
+        Merge two OID-to-profile mappings without modifying either input.
+
+        Each dictionary maps an OID to one or more comma-separated profile
+        names. When an OID exists in both dictionaries, each distinct profile
+        name is retained once, in first-seen order.
+
+        The previous logic compared the full comma-separated profile string,
+        so duplicate profiles could be added when the profile order differed.
+
+        :param dict1: Base OID-to-profile mapping.
```
omrozowicz-splunk: maybe let's use this PR to change dictN into some meaningful name 😂

omrozowicz-splunk (commented):
I wonder if merging and dividing string comma separated list is really necessary, maybe would be better to leave it as lists? 🤔 but anyway, this is a fix and it works so this version also works

splunk_connect_for_snmp/snmp/varbinds_resolver.py
```diff
@@ -221,20 +221,62 @@ def __repr__(self):
...
+            dict2_profiles = self.unique_profile_names(profile_names)
+
+            dict1_profile_set = set(dict1_profiles)
+            added_profiles = [
+                profile_name
+                for profile_name in dict2_profiles
+                if profile_name not in dict1_profile_set
+            ]
+            result_profiles = dict1_profiles + added_profiles
+            mapping[mapping_key] = ",".join(result_profiles)
```
ajasnosz: Do we need to split and join the dict at every iteration? Wouldn't it be better to join at the end outside the loop, once everything is processed?

splunk_connect_for_snmp/snmp/varbinds_resolver.py
```diff
@@ -221,20 +221,62 @@ def __repr__(self):
...
-                mapping[k] = f"{mapping[k]},{v}"
-            elif not mapping.get(k):
-                mapping[k] = v
+        mapping = {
+            mapping_key: ",".join(self.unique_profile_names(profile_names))
+            for mapping_key, profile_names in dict1.items()
+        }
+
+        for mapping_key, profile_names in dict2.items():
+            dict1_profiles = self.unique_profile_names(mapping.get(mapping_key, ""))
```
ajasnosz: if unique is static method shouldn't we use it as Profile.unique_profile_names?

splunk_connect_for_snmp/snmp/varbinds_resolver.py
gmaiyani: > can we name the dic1 and dict2 something more meaningful like base / incoming mapping?

Yes sure, updated in latest commit.

splunk_connect_for_snmp/snmp/varbinds_resolver.py
gmaiyani: Nope, it would be good to add later outside of the loop, updated in the latest commit.

splunk_connect_for_snmp/snmp/varbinds_resolver.py
gmaiyani: > if unique is static method shouldn't we use it as Profile.unique_profile_names?

Yes that's the other way we can use. Ive updated accordingly.

omrozowicz-splunk (approved):

ajasnosz (approved):

PR #1469, opened 2026-07-27
chore: run one poll & walk within one event loop owned snmp engine

gmaiyani:
# Description
- Run single poll and walk within the one event loop owned snmp engine.
- Provided support of nonRepeaters in multi_bulk_walk_cmd

Fixes # (issue)

## Type of change

- [x] Refactor/improvement

## How Has This Been Tested?

## Checklist

### DIFF
--- a/splunk_connect_for_snmp/snmp/manager.py
+++ b/splunk_connect_for_snmp/snmp/manager.py
@@ -413,6 +413,11 @@ class Poller(Task):
             address, is_walk=is_walk, profiles=profiles
         )
 
+        metrics: Dict[str, Any] = {}
+        if not varbinds_get and not varbinds_bulk:
+            logger.info(f"No work to do for {address}")
+            return False, {}
+
         # Keeping the engine in this scope closes its transports before the poll or
         # walk leaves the event loop pysnmp bound them to.
         with SnmpEngine() as snmp_engine:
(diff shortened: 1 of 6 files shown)


splunk_connect_for_snmp/snmp/manager.py
```diff
@@ -78,6 +78,14 @@
     "UDP-MIB",
 ]
 
+PYSNMP_PROTOCOL_MIBS = (
+    "SNMPv2-MIB",
```
omrozowicz-splunk, line 82: SNMPv2-MIB is also in DEFAULT_STANDARD_MIBS, is it intended?

splunk_connect_for_snmp/snmp/manager.py
```diff
@@ -414,45 +413,53 @@ async def do_work(
...
 
-        metrics: Dict[str, Any] = {}
-        if not varbinds_get and not varbinds_bulk:
-            logger.info(f"No work to do for {address}")
-            return False, {}
+            transport = await setup_transport_target(ir)
 
-        max_oid = ir.max_oid_to_process if ir.max_oid_to_process else MAX_OID_TO_PROCESS
+            metrics: Dict[str, Any] = {}
+            if not varbinds_get and not varbinds_bulk:
```
omrozowicz-splunk: Can we put it before opening the SnmpEngine context manager? Looks like this check doesn't need any network connection

splunk_connect_for_snmp/snmp/manager.py
gmaiyani: Yes, it is intended to keep the PYSNMP & DEFAULT_STANDARD_MIBS separate. it would not load the mib twice into the same builder. So we can keep it as it is.

omrozowicz-splunk (approved):

PR #1473, opened 2026-08-04
fix: revalidate cached MIB index on worker startup

gmaiyani:
# Description

Revalidates the Mongo-backed MIB index cache whenever a trap or poller worker starts/restart. This ensures newly compiled local MIBs are discovered immediately after the relevant worker is restarted, without waiting for the 30-minute cache expiry or manually clearing MongoDB cache.

more details can be found here: https://github.com/splunk/splunk-connect-for-snmp/discussions/1459

Fixes # (issue)
- Revalidates the Mongo-backed MIB index cache whenever a trap or poller worker starts/restart. 

## Type of change

- [x] Bug fix

## How Has This Been Tested?

## Checklist

### DIFF
--- a/docs/dockercompose/7-docker-logging.md
+++ b/docs/dockercompose/7-docker-logging.md
@@ -13,6 +13,22 @@ DOCKER_LOG_COMPRESS=true
 
 Docker applies logging-driver changes when containers are created. Run `docker compose up -d` after changing the logging configuration so Compose can recreate the affected containers.
 
+## Prerequisites
+
+Running `manage_logs.py` requires the `ruamel.yaml` Python package. Install it with:
+
+```shell
+pip3 install ruamel.yaml
+```
+
+## `manage_logs.py` flags
+
+| Flag                                  | Description                                                                                     |
+|---------------------------------------|-------------------------------------------------------------------------------------------------|
+| `-p`, `--path_to_compose`             | Absolute path to the directory containing the Docker Compose files                              |
+| `--configure_default_logging`         | Apply bounded `json-file` logging or Docker daemon inheritance based on `DOCKER_LOG_DRIVER`     |
+| `--use_docker_default_logging`        | Remove service-level logging configuration so containers inherit the Docker daemon configuration |
+
 ## Choose the logging behavior
 
 Check the Docker daemon's current logging driver before a new installation or upgrade:
--- a/docs/dockercompose/7-docker-logging.md
+++ b/docs/dockercompose/7-docker-logging.md
@@ -1,0 +1,106 @@
+# Docker logging
+
+SC4SNMP applications write logs to container standard output and standard error. Docker controls how those container logs are stored and retained. These settings do not change SC4SNMP log levels, messages, formats, or destinations.
+
+The default Compose configuration uses the `json-file` driver with bounded rotation for every service:
+
+```dotenv
+DOCKER_LOG_DRIVER=json-file
+DOCKER_LOG_MAX_SIZE=10m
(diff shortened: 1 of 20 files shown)


docs/dockercompose/7-docker-logging.md
```diff
@@ -13,6 +13,22 @@ DOCKER_LOG_COMPRESS=true
 
 Docker applies logging-driver changes when containers are created. Run `docker compose up -d` after changing the logging configuration so Compose can recreate the affected containers.
 
+## Prerequisites
```
omrozowicz-splunk, line 16: this looks like unrelated part of another MR, no?

splunk_connect_for_snmp/snmp/manager.py
```diff
@@ -320,6 +398,7 @@ def __init__(self, **kwargs):
 
         if kwargs.get("no_mongo"):
```
omrozowicz-splunk, line 322: is this `no_mongo` parameter actually used anywhere? it looks like a leftover

splunk_connect_for_snmp/snmp/manager.py
```diff
@@ -166,6 +168,82 @@ def is_mib_resolved(id):
         return True
 
 
+def _parse_mib_index(index_content: str) -> Tuple[Dict[str, str], int]:
```
omrozowicz-splunk: this is called twice, once in the `Poller` task in `_refresh_mib_map` and second time in `_validate_mib_index_response`. Can we just run it once and save the results instead?

omrozowicz-splunk (commented):
maybe let's add unit tests for the new methods? it looks quite easy to cover with tests

docs/dockercompose/7-docker-logging.md
gmaiyani: Yes, i forgot to mention in previous PR so i kept it in this one

splunk_connect_for_snmp/snmp/manager.py
gmaiyani: i guess not. we have one script called `walk.py` (which uses that no_mongo) but I'm not sure if we are using it anymore. 

here: https://github.com/splunk/splunk-connect-for-snmp/blob/4210dc59bea9867783a02ab66745ff9d6c5299b5/splunk_connect_for_snmp/walk.py#L34

Well no one touched it in last 4 years so 😂

splunk_connect_for_snmp/snmp/manager.py
gmaiyani: both are used in different cases.
1. `_validate_mib_index_response()` parses the live response before requests cache stores it. It will prevents an invalid live index from replacing the last  valid cached response
2. `_refresh_mib_map()` parses the final response which was returned by the session. It may be a live response or a cached response confirmed through 304 revalidation, or a stale cached fallback. Keeping this final parse ensures that mib_map is always built from the response actually returned to the worker. For a valid live response this can duplicate the hooks validation parse, but the refresh currently runs only during Poller initialization, so would not cause the issue.

Let me check if its better to used request-scoped validator instead of this may be

splunk_connect_for_snmp/snmp/manager.py
omrozowicz-splunk: okay, I was just wondering what it was 😂 let's plan to see and assess if we can clean it up

gmaiyani: Sure :)

omrozowicz-splunk (approved):

PR #1494, opened 2026-09-03
fix: preserve replica key across upgrades and support no-auth to auth…

gmaiyani:
# Description

- Added pre-upgrade and post-upgrade Jobs to coordinate the MongoDB authentication transition.
- Removed the unused mongodb.auth.existingReplicaKeySecret option.
- Standardized the replica-key Secret name as `<release-name>-mongodb-replicakey`.
- Added mongodb.replicationAuthTransition.transitionTimeout to configure the transition Job timeout.
- Added clearer progress and failure logs for each authentication-transition phase.
- Included UI backend and UI worker Deployments in MongoDB client rollout checks when the UI is enabled.
Updated MongoDB configuration, upgrade, troubleshooting, and changelog documentation.

Fixes # (issue)

- Prevented the MongoDB replica key from being regenerated during every Helm upgrade.
- Reused the existing `<release-name>-mongodb-replicakey `Secret value across upgrades.
- Added a safe transition from unauthenticated to authenticated MongoDB replication using transitionToAuth.
- Added ordered MongoDB member restarts, restarting SECONDARY members before stepping down and restarting the PRIMARY.
- Ensured SC4SNMP services switch to authenticated MongoDB connections before strict authentication is enforced.
- Added support for resuming an interrupted authentication transition.

## Type of change

- [x] Bug fix
- [x] Refactor/improvement

## How Has This Been Tested?

manually

## Checklist


charts/splunk-connect-for-snmp/templates/mongodb/mongodb-enable-auth-rollout-job.yaml
```diff
@@ -0,0 +1,367 @@
...
+      serviceAccountName: {{ .Release.Name }}-mongodb-init-sa
+      securityContext:
+        runAsNonRoot: true
+        runAsUser: 65534
+        runAsGroup: 65534
+        seccompProfile:
+          type: RuntimeDefault
+      containers:
+      - name: migrate-replica-set
+        image: {{ .Values.mongodb.replicaInitJob.image.repository | default "alpine/kubectl" }}:{{ .Values.mongodb.replicaInitJob.image.tag | default "1.36.3" }}
```
ajasnosz: not sure if we need to keep the defaults here, I think it should be always tracked from variables, otherwise we are adding another point to keep track when version bumping

charts/splunk-connect-for-snmp/templates/mongodb/mongodb-enable-auth-rollout-job.yaml
```diff
@@ -0,0 +1,367 @@
...
+                fi
+                candidate=${target}
+                break
+              done
+
+              if [ -n "${candidate}" ]; then
+                if [ "$(get_member_role "${candidate}" "${phase_mode}")" != "SECONDARY" ]; then
+                  echo "[ERROR] ${candidate} is not a healthy SECONDARY" >&2
+                  return 1
+                fi
```
ajasnosz: wouldn't it be better to also use here wait_for_member before checking the secondary?

docs/microk8s/configuration/mongo-configuration.md
```diff
@@ -308,6 +317,66 @@ To rotate the password without losing data, perform the steps **in this order**:
...
+!!!warning "Wait for the inventory Job to be removed"
+    The inventory Job has an immutable Pod template. Enabling MongoDB authentication changes that template to include the MongoDB credentials. If the previous inventory Job still exists, the upgrade fails with `spec.template: field is immutable`.
+
+    The inventory Job can be created during installation or configuration updates, including when **Apply changes** is selected in the UI. Before starting the authentication upgrade, wait for the Job to complete, then wait until `<release-name>-splunk-connect-for-snmp-inventory` is no longer listed by `microk8s kubectl get jobs --namespace sc4snmp`.
+
+The transition can take longer than Helm's default timeout. When changing authentication from disabled to enabled, use an additional timeout for the upgrade:
+
+```yaml
+microk8s helm3 upgrade --install snmp -f values.yaml splunk-connect-for-snmp/splunk-connect-for-snmp --namespace=sc4snmp --create-namespace --timeout 30m
+```
```
ajasnosz: We are suggesting here the timeout as 30m, and even if it is a suggestion the customers following the docs can just copy that value, but our pre-upgrade and post-upgrade jobs timeouts sums up to 42m of waiting, so I think we can add here the recommendation to set the timeout as (time from pre-upgrade job + time from post-upgrade job + some buffer)

CHANGELOG.md
```diff
@@ -3,12 +3,15 @@
...
 ### Changed
+- Add `mongodb.replicationAuthTransition.transitionTimeout` to configure the authentication-transition deadline, which defaults to one replica initialization timeout per MongoDB member.
 - Add configurable IPv6 listener support for Redis, MongoDB, and the MIB server with `redis.ipv6Enabled`, `mongodb.ipv6Enabled`, and `mibserver.ipv6Enabled`.
 - Extend the existing Docker Compose `IPv6_ENABLED` option to configure MongoDB and MIB server listeners.
 - Provided `MAX_MEMORY_PER_CHILD` / `maxMemoryPerChild` setting to celery workers.
 - Provided configurable options to control the docker logging: `DOCKER_LOG_DRIVER`, `DOCKER_LOG_MAX_SIZE`, `DOCKER_LOG_MAX_FILE`, `DOCKER_LOG_COMPRESS` to prevent the disk usage growth
 
 ### Fixed
+- Preserve the chart-managed MongoDB replica key across Helm upgrades and remove the unused option for an external replica-key Secret.
+- Support enabling authentication on an existing MongoDB replica set with MongoDB's `transitionToAuth` compatibility mode, ordered member replacement, resumable phases, and verification of MongoDB members and SC4SNMP clients.
```
ajasnosz: it should be one entry per pr in one section, if there is more steps to describe let's add another intent level. 
Like with trap handling improvements in 1.17.0

docs/microk8s/configuration/mongo-configuration.md
```diff
@@ -167,6 +173,9 @@ mongodb:
   replicaSetName: rs0
 ```
 
+!!!note "MongoDB replica key"
```
ajasnosz: maybe it's more of an warning?

Dockerfile
```diff
@@ -17,6 +17,7 @@ COPY splunk_connect_for_snmp /app/splunk_connect_for_snmp
 WORKDIR /app
 RUN poetry config virtualenvs.in-project true ;\
     poetry install --no-root --sync --without dev ;\
+    poetry install --only main --no-root --no-interaction ;\
```
ajasnosz: line 20 got merged incorrectly, it shouldn't be there anymore, only line 19

Dockerfile
```diff
@@ -3,7 +3,7 @@ FROM python:3.10-alpine AS base
 ENV PYTHONFAULTHANDLER=1 \
     PYTHONHASHSEED=random \
     PYTHONUNBUFFERED=1
-RUN apk add --no-cache --upgrade git sqlite-dev xz-libs libcrypto3 libssl3
+RUN apk upgrade --no-cache && apk add --no-cache git sqlite-dev xz-libs libcrypto3 libssl3
```
ajasnosz, line 6: not sure about change in this line, if it's also some wrongly resolved rebase from develop, or it should be changed like this?

ajasnosz (changes requested):

Dockerfile
gmaiyani: Yes, accidentally added incoming + current changes 😅

Dockerfile
gmaiyani: this change is intended to fix the trivy failure

charts/splunk-connect-for-snmp/templates/mongodb/mongodb-enable-auth-rollout-job.yaml
```diff
@@ -0,0 +1,367 @@
...
+  template:
+    metadata:
+      labels:
+        app: {{ .Release.Name }}-mongodb-auth-migration
+    spec:
+      restartPolicy: Never
+      serviceAccountName: {{ .Release.Name }}-mongodb-init-sa
+      securityContext:
+        runAsNonRoot: true
+        runAsUser: 65534
```
omrozowicz-splunk, line 41: Is it `65534` because it is hardcoded somewhere else in mongodb already? 🤔

omrozowicz-splunk:
I see the change is pretty big, how we will make sure it is not broken after we update the mongo version? I think we should have regression tests that are run automatically in the CI so we're sure all those commands are still valid.

omrozowicz-splunk:
also, will we need the same for redis?

charts/splunk-connect-for-snmp/templates/mongodb/mongodb-enable-auth-rollout-job.yaml
gmaiyani: nope, I think `65534` is the standard uid for the unprivileged nobody user, and it works here because this Job uses only the alpine/kubectl image and runs kubectl. So, `65534` is a better.

charts/splunk-connect-for-snmp/templates/mongodb/mongodb-enable-auth-rollout-job.yaml
gmaiyani: yes, I'll check the other files as well and make the changes there too.

charts/splunk-connect-for-snmp/templates/mongodb/mongodb-enable-auth-rollout-job.yaml
gmaiyani: yes, I think that is better than checking the state only. we should wait for a specific time for the member to become healthy.

docs/microk8s/configuration/mongo-configuration.md
gmaiyani: yess

gmaiyani:
> also, will we need the same for redis?

ahh, i almost forgot the redis 🫠. I will check.

gmaiyani:
> I see the change is pretty big, how we will make sure it is not broken after we update the mongo version? I think we should have regression tests that are run automatically in the CI so we're sure all those commands are still valid.

may be let's add the integration test cases for this ? it would be better i think 🤔.

ajasnosz (approved):

entrypoint.sh
```diff
@@ -20,7 +20,7 @@ wait-for-dep ${REDIS_DEPENDENCIES} "${MONGO_WAIT}" "${MIB_INDEX}"
 ENABLE_TRAPS_SECRETS=${ENABLE_TRAPS_SECRETS:=false}
 ENABLE_WORKER_POLLER_SECRETS=${ENABLE_WORKER_POLLER_SECRETS:=false}
 ENABLE_WORKER_DISCOVERY_SECRETS=${ENABLE_WORKER_DISCOVERY_SECRETS:=false}
-wait-for-dep "${REDIS_DEPENDENCIES}" "${MONGO_URI}" "${MIB_INDEX}"
+wait-for-dep ${REDIS_DEPENDENCIES} "${MONGO_URI}" "${MIB_INDEX}"
```
ajasnosz, line 23: why do we remove the "" only for redis?

gmaiyani: It’s because `REDIS_DEPENDENCIES` is a space-separated list of multiple urls. In replication mode, there are four redis/sentinel dependencies, while in standalone mode, there are only two redis dependencies. wait-for-dep expects each url as a separate argument, but since the variable was in quotes, all the url were treated as a single argument, causing it to wait on an invalid url. In standalone mode, it worked even though the urls were quoted because the redis URL parser accepted the combined value as the first endpoint and effectively ignored the remaining url.

At line 18, we are already using it correctly, but later we accidentally added the quotes 🫠

